PR #524, opened 2021-11-15
[CP-827] The no result state isn't display when user doesn't have messages

dkarski:
Jira: [CP-827]

**Description**

EmptyState component added to Messages Page when threads length size is equal 0

[CP-827]: https://appnroll.atlassian.net/browse/CP-827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/app/src/messages/components/messages/messages-test-ids.enum.ts
+++ b/packages/app/src/messages/components/messages/messages-test-ids.enum.ts
@@ -4,6 +4,8 @@
  */
 
 export enum MessagesTestIds {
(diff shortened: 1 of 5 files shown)


igorbezsmertnyi (approved):

wercix (approved):
Maybe we should add a loading state as the next iteration? 🤔

PR #593, opened 2022-01-17
[CP-961] Implement `getData` method `BaseIndexer` based on backuping process

dkarski:
Jira: [CP-961]

**Description**

[CP-961]: https://appnroll.atlassian.net/browse/CP-961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wercix (approved):

packages/app/src/data-sync/indexes/base.indexer.ts
```diff
@@ -3,19 +3,43 @@
...
-      locateFile: (fileName: string) =>
-        `${__dirname}/../static/${fileName}`,
+      locateFile: (fileName: string) => `${__dirname}/../static/${fileName}`,
     })
   }
 
-  public async getData(dbFileName: string): Promise<ArrayLike<number> | Buffer | null> {
-    // TODO: Will be Implement in the next subtask -> CP-961
-    return null
+  public async getData(
```
igorbezsmertnyi, line 26: I'm not sure that this is a good idea. Let's imaging that we have two indexers, in this case, when we will try to exec the `getData` method on each of them it will initiate backup download twice. I suggest moving the logic of the downloading and extracting to the top-level instance and leaving here only the getting file by name or something.

igorbezsmertnyi (changes requested):

igorbezsmertnyi (approved):

wercix (approved):

PR #638, opened 2022-02-13
[CP-1015] I would like to know that the Initializing Connection can takes more time

wercix:
Jira: [CP-1015]

**Description**

- changed description if connecting will be longer
- add error modal for sync problems

<details>

Screen Shot 2022-02-13 at 20 07 35

Screen Shot 2022-02-13 at 20 22 40

</details>

[CP-1015]: https://appnroll.atlassian.net/browse/CP-1015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/app/src/connecting/components/connecting.component.tsx
```diff
@@ -108,17 +117,30 @@ const Connecting: FunctionComponent<{
...
+  const onRetry = () => {
+    // TODO: do some logic to retry connection
+    history.push(URL_ONBOARDING.connecting)
+  }
+
   return (
     <>
-      {error && <ErrorConnectingModal open closeModal={close} />}
+      {error &&
+        (syncInitialized ? (
```
dkarski, line 128: when the sync error is thrown then the `syncInitialized` cannot be true, so this condition doesn't run ever.
```suggestion
        (syncState === SynchronizationState.Error ? (
```

wercix: But I changed the condition of setting error and now it doesn't depend on _syncInitialized_
`useEffect(() => {
    if (unlocked && syncState === SynchronizationState.Error) {
      setError(true)
    }
  }, [syncInitialized, syncState, unlocked])`

dkarski: I know, but even then 😛 because when the sync error is thrown, the reducer no set the `syncInitialized` to true 🙃

wercix: If I change the condition as you suggested then `ErrorConnectingModal` will be never shown because `syncState === SynchronizationState.Error`  is also a condition for error state

wercix: Added error types for better conditioning in [b5ac601](https://github.com/mudita/mudita-center/pull/638/commits/b5ac601229195b7cd4decca8e7b1a48816b6b95c)

dkarski (dismissed):

dkarski (commented):
Are you sure that it is needed? 😬 
Zrzut ekranu 2022-02-15 o 10 42 29

wercix:
> Are you sure that it is needed? 😬 Zrzut ekranu 2022-02-15 o 10 42 29

It came from master so I think it is needed. It is in a separate commit because I misclicked it before.   🤷🏻‍♀️

dkarski:
> > Are you sure that it is needed? 😬 Zrzut ekranu 2022-02-15 o 10 42 29
> 
> It came from master so I think it is needed. It is in a separate commit because I misclicked it before. 🤷🏻‍♀️

Ok, I see... It just a prettier 😅 🤦 

Zrzut ekranu 2022-02-15 o 10 51 53

dkarski (approved):

igorbezsmertnyi (approved):

PR #703, opened 2022-04-21
[CP-1239] Fix Messages view after removed thread on the Pure

wercix:
Jira: [CP-1239]

**Description**

<details>
// put images here
</details>

[CP-1239]: https://appnroll.atlassian.net/browse/CP-1239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/app/src/messages/components/messages/messages.component.tsx
+++ b/packages/app/src/messages/components/messages/messages.component.tsx
@@ -269,6 +269,8 @@ const Messages: FunctionComponent<Props> = ({
       )
       if (thread) {
         openThreadDetails(thread)
+      } else {
+        setActiveThread(undefined)
       }
     }
   }, [activeThread, threads])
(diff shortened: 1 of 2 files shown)


dkarski (approved):

igorbezsmertnyi:
@wercix Here should be the same mechanic as there https://github.com/mudita/mudita-center/pull/704#issuecomment-1106098988

wercix:
> @wercix Here should be the same mechanic as there [#704 (comment)](https://github.com/mudita/mudita-center/pull/704#issuecomment-1106098988)

I don't think so. This is different case. Here we have only UI problem. Showing thread details depends only on Messages component states (messagesState and activeThread) and there will be no other change in the data model, only thread delete as it is.

igorbezsmertnyi (approved):

PR #750, opened 2022-06-02
[CP-526] Join template creation logic

igorbezsmertnyi:
Jira: [CP-526]

**Description**
Connection template form with the main process via redux

<details>
// put images here
Screen Shot 2022-06-03 at 12 46 43 PM

</details>

[CP-526]: https://appnroll.atlassian.net/browse/CP-526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dkarski (approved):

packages/app/src/templates/actions/create-template.action.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  lastUsedAt: "2",
+}
+
+beforeEach(() => {
+  mockStore.clearActions()
+})
+
+jest.mock("App/templates/requests/create-template.request")
+
+describe("async `createTemplate` ", () => {
```
wojcik92michal, line 35:
```suggestion
describe("async `createTemplate`", () => {
```

packages/app/src/templates/actions/create-template.action.test.ts
```diff
@@ -0,0 +1,102 @@
...
+      expect(mockStore.getActions()).toEqual([
+        createTemplate.pending(requestId, newTemplate),
+        createTemplate.fulfilled(template, requestId, newTemplate),
+      ])
+
+      expect(createTemplateRequest).toHaveBeenCalled()
+    })
+  })
+
+  describe("when `createTemplateRequest` return undefine `data` or `error` exists", () => {
```
wojcik92michal, line 58:
```suggestion
  describe("when `createTemplateRequest` return undefined `data` or `error` exists", () => {
```

packages/app/src/templates/actions/create-template.action.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    test("fire async `createTemplate` returns `rejected` action if data is undefine", async () => {
```
wojcik92michal, line 59:
```suggestion
    test("fire async `createTemplate` returns `rejected` action if data is undefined", async () => {
```

wojcik92michal (approved):

packages/app/src/templates/reducers/template.interface.ts
```diff
@@ -3,9 +3,26 @@
...
 
+import { PayloadAction } from "@reduxjs/toolkit"
+import { TemplatesEvent } from "App/templates/constants"
+import { CreateTemplateError } from "App/templates/errors"
 import { Template } from "App/templates/dto"
 
 export interface TemplateState {
   data: Template[]
+  loading: boolean
+  loaded: boolean
```
wercix, line 14: Do we really need separate loaded and loading state? maybe we can add a state `addingNewTemplateState` which will be enum like
`export enum addingNewTemplateState {
  Loading,
  Loaded,
  Empty,
  Error,
}`
Current names can be confusing because there will be more loading states (for example for editing and deleting).

packages/app/src/templates/reducers/template.interface.ts
```diff
@@ -3,9 +3,26 @@
  * For licensing, see https://github.com/mudita/mudita-center/blob/master/LICENSE.md
  */
 
+import { PayloadAction } from "@reduxjs/toolkit"
+import { TemplatesEvent } from "App/templates/constants"
+import { CreateTemplateError } from "App/templates/errors"
 import { Template } from "App/templates/dto"
 
 export interface TemplateState {
   data: Template[]
```
wercix, line 12: I'm thinking, maybe here `data` is too generic name? 
```suggestion
  templates: Template[]
```

igorbezsmertnyi: According to our style guild, the `data` key is used in each reducer. It makes reducers more generic and undrestandeble.

packages/app/src/templates/reducers/template.interface.ts
igorbezsmertnyi: Currently, we will not support two types of action in MC. If you, for example, add a new record, you can't edit anything simultaneously, so this action should be locked.

dkarski (approved):

wojcik92michal (approved):

wercix (approved):

PR #770, opened 2022-06-09
[CP-964] I want delete my template.

wercix:
Jira: [CP-964]

**Description**
Sum of PRs connected to delete one template. 
<details>
<summary><b>Video</b></summary>

https://user-images.githubusercontent.com/15948014/172802510-93ae5809-3cec-4664-91be-d2dfdad7b920.mov

</details>

[CP-964]: https://appnroll.atlassian.net/browse/CP-964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/app/src/backend/device-service.ts
```diff
@@ -149,6 +149,11 @@ export class DeviceService {
     method: Method.Delete
     body: GetThreadBody
   }): Promise<RequestResponse>
+  public request(config: {
+    endpoint: Endpoint.Messages
+    method: Method.Delete
+    body: GetTemplateBody
```
igorbezsmertnyi, line 155: It's a delete method, so let's use the `DeleteTemplateBody`

packages/app/src/templates/reducers/template.reducer.ts
```diff
@@ -4,25 +4,27 @@
...
   fulfilledAction,
 } from "Renderer/store/helpers"
-import { TemplatesEvent } from "App/templates/constants"
 
 export const initialState: TemplateState = {
   data: [],
   loaded: false,
   loading: false,
   error: null,
+  deletingState: null,
```
igorbezsmertnyi, line 27: I remember that we discussed not using the extra states. You can check the status of the previous operation by comparing existing states.
Example:
```ts
Error: loaded === false && loading === false && error !== null
Progress: loaded === false && loading === true && error === null
Success: loaded === true && loading === false && error === null
```

wercix: Yes, but I also need information that loaded, loading or error state is connected to deleting, not to add or editing template. So in this case I still need another state for example `deleting: boolean` 
Error: loaded === false && loading === false && error !== null && deleting 
Progress: loaded === false && loading === true && error === null && deleting 
Success: loaded === true && loading === false && error === null && deleting

igorbezsmertnyi: In each case responsible for deleting you can update the `loading/loaded` state too

dkarski (dismissed):

packages/app/src/backend/device-service.ts
wercix: There was already a proper request. I removed this one [f9d5dde](https://github.com/mudita/mudita-center/pull/770/commits/f9d5ddea9864fd02e95e471b3e5f9c2f4e3b4d8a)

packages/app/src/templates/reducers/template.reducer.ts
wercix: So, I removed deletingState, but added boolean deleting as in suggestion [02d3381](https://github.com/mudita/mudita-center/pull/770/commits/02d33817e550d87c036c35fb46efc745904b1e57)

igorbezsmertnyi (approved):

PR #787, opened 2022-06-20
[CP-1361] Fix page titles setting functionality

igorbezsmertnyi:
Jira: [CP-1361]

**Description**
Resolve page `title` issue by removing title tag from HTML template

<details>
// put images here
</details>

[CP-1361]: https://appnroll.atlassian.net/browse/CP-1361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/app/templates/index.html
```diff
@@ -0,0 +1,8 @@
+<!DOCTYPE html>
```
wojcik92michal, line 1: Hmm, I couldn't find the `title` tag generated. Is it ok? Do the e2e tests work fine (e.g. `sar-window-check.e2e.ts`)? 

image

igorbezsmertnyi: Yes, I know. But, it's works in e2e test

wojcik92michal: 😮 

Just curious, do you know why? Previously, when I was simply setting the `title` tag in `BrowserWindow`, it was not working. Maybe the reason is that the title tag is not generated at all? 

image

wojcik92michal (approved):

packages/app/templates/index.html
igorbezsmertnyi: Yep, it was overridden [here](https://github.com/mudita/mudita-center/pull/787/files#diff-8b54cdec43a0d5fe62f1f0022b422d3ba33eba5b41dd63638736e7ce7fed12abL42) by HTMLPlugin which put an empty `title` tag to HTML. That's why you should manually set the title to each page. But now, the electron automatically coupe with it.

dkarski (approved):

PR #820, opened 2022-07-01
[CP-1281] improve mudita news performance

wojcik92michal:
Jira: [CP-1281]

**Description**

<details>
// put images here
</details>


wercix (dismissed):
How about loading state for this starting 1 second? 🤔

wojcik92michal:
> How about loading state for this starting 1 second? 🤔

@wercix 
At the beginning of the implementation phase I thought about adding it, but after the discussion with @dkarski we decided not to add it due to some potential delays (it would need additional work + there are "delaying" factors like waiting for designs, requirements etc.) and to take the incremental approach (if needed, we would add it later).

Is this approach ok for you?

igorbezsmertnyi (approved):

wercix (approved):

PR #932, opened 2022-08-24
[CP-1555] User can't update Center in the Settings/About section

dkarski:
Jira: [CP-1555]

**Description**

### DIFF
--- a/packages/app/src/settings/reducers/settings.reducer.ts
+++ b/packages/app/src/settings/reducers/settings.reducer.ts
@@ -18,7 +18,6 @@ import {
   toggleTethering,
   toggleUpdateAvailable,
   toggleCollectionData,
-  checkUpdateAvailable,
   setConversionFormat,
   setConvert,
   setNonStandardAudioFilesConversion,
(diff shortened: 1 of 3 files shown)


igorbezsmertnyi (approved):

packages/app/src/settings/reducers/settings.reducer.ts
```diff
@@ -97,24 +96,14 @@ export const settingsReducer = createReducer<SettingsState>(
...
 
-      .addCase(toggleUpdateAvailable.pending, (state) => {
-        state.loaded = false
-        state.loading = true
-      })
-      .addCase(toggleUpdateAvailable.fulfilled, (state, action) => {
-        state.loaded = true
-        state.loading = false
-        state.updateAvailable = action.payload
+      .addCase(toggleUpdateAvailable.pending, (state, action) => {
```
wojcik92michal, line 99: Is it ok that this action is handled in `pending` state, not in `fulfilled`?

wercix (approved):

packages/app/src/settings/reducers/settings.reducer.ts
dkarski: I think that changing the state during the pending state isn't bad. Maybe, the architecture of the action definitions could be better here but please leave this improvement for another chance.

PR #952, opened 2022-09-01
[CP-1523] Improve phone number regexp

igorbezsmertnyi:
Jira: [CP-1523]

**Description**
Improve phone number validation

<details>
// put images here
</details>

[CP-1523]: https://appnroll.atlassian.net/browse/CP-1523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/app/src/__deprecated__/renderer/utils/form-validators.ts
+++ b/packages/app/src/__deprecated__/renderer/utils/form-validators.ts
@@ -6,7 +6,8 @@
 import { RegisterOptions } from "react-hook-form/dist/types"
 import { intl } from "App/__deprecated__/renderer/utils/intl"
 
-export const phoneNumberRegexp = /^[ \\.+\-#()\d]*$/im
+export const phoneNumberRegexp =
+  /^\+?\d{1,4}?[-.\s]?\(?\d{1,3}?\)?[-.\s]?\d{1,4}[-.\s]?\d{1,4}[-.\s]?\d{1,9}$/
 
 export const passwordRegexp =
   /^(?=.*\d)(?=.*[!"#$%&'()*+,-./:;<=>?@[\]^_`{|}~])(?=.*[a-z])(?=.*[A-Z]).{8,}$/


packages/app/src/__deprecated__/renderer/utils/form-validators.ts
```diff
@@ -6,7 +6,8 @@
 import { RegisterOptions } from "react-hook-form/dist/types"
 import { intl } from "App/__deprecated__/renderer/utils/intl"
 
-export const phoneNumberRegexp = /^[ \\.+\-#()\d]*$/im
+export const phoneNumberRegexp =
```
wojcik92michal, line 9: Any tests for that? Regexp are tricky?

packages/app/src/__deprecated__/renderer/utils/form-validators.ts
```diff
@@ -6,7 +6,8 @@
...
+  /^\+?\d{1,4}?[-.\s]?\(?\d{1,3}?\)?[-.\s]?\d{1,4}[-.\s]?\d{1,4}[-.\s]?\d{1,9}$/
```
wojcik92michal, line 10: Seems to me It won't match numbers like `997` or `*101#` (checking account state in Play)

Phone number regex is always tricky and problematic. Maybe we should:
1. use the lib recommended from https://stackoverflow.com/a/15644461?
2. trust users that they can put the correct number and allow only numbers, `#` or `*` in any order (so that e.g. `*555*#123*` is allowed even if it is not the correct number)? This is the approach taken by iPhone that I would recommend as well (but not sure if it would match our business needs).

igorbezsmertnyi: @wojcik92michal Unfortunately it's impossible to use an external library. Each of those libraries required a county code to validate the phone numbers

wojcik92michal: Ah, I see. So maybe we should consider second point? Or maybe apply the same validation rules as in Pure device (not sure how they look like)?

igorbezsmertnyi: @wojcik92michal Pure, use some `c++` which resolve those issues for them. But unfortunately there is no port for JS

igorbezsmertnyi: @wojcik92michal I see two solutions here: 
- Check if the pure can give us the information about the operator country or something similar
- Make a more generic regexp

wojcik92michal: Sure - I would opt for more generic regexp. But I think it would be good to confirm that with business or requirements.

wercix: Please check LibPhoneNumber library that is used on the Pure side, maybe it will be a good solution ([here](https://www.npmjs.com/package/libphonenumber-js) is js version)

igorbezsmertnyi: @wercix 
> @wojcik92michal Unfortunately it's impossible to use an external library. Each of those libraries required a county code to validate the phone numbers

As I mentioned above, this library required county code to validate phone numbers

wojcik92michal: @igorbezsmertnyi Interesting fact - in theory this lib it does not require country code to be passed directly, but the validation is not the best.

image

dkarski:
I am closing this PR until the refinement of the issue will be done.

PR #1057, opened 2022-12-01
[Cp-1738] integrate with get release endpoint changes

wojcik92michal:
Jira: [CP-1738]

**Description**
Scope:
* update `update-available-modal.component.tsx` so that it uses shared components from our codebase (instead of raw library components)
* integrating with get-release-v2 endpoint

Notes:
* downloading/updating will happen only for latest release - those features are scope of another tasks

Testing notes:
* process should work for Harmony/Pure
* to test the flow for Pure install OS 1.2.0 (to see few possible updates) or 1.3.0 (to see single update)
* to test the flow for Harmony install OS 1.5.0 (to see few possible updates) or 1.6.0 (to see single update)
* to trigger the dialog go to "Overview" page and quickly click for "Check for update" (before it turns into "Download now"). In the background, there is a silent process that checks if there are releases for update. The behaviour of the button needs be discussed and settled with business an designer. 

<details>
// put images here
</details>

[CP-1738]: https://appnroll.atlassian.net/browse/CP-1738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/app/src/overview/components/update-os-modals/update-available-modal/update-available-modal.component.tsx
```diff
@@ -53,47 +39,16 @@ const messages = defineMessages({
...
-      onAfterClose={onClose}
-    >
-      <Header
-        titleOrder={TitleOrder.TitleFirst}
-        subtitleGap={false}
-        data-testid={ModalTestIds.Header}
+export const UpdateAvailableModal: FunctionComponent<UpdateAvailableModalProps> =
+  ({ open = false, releases, onClose, onDownload, testId }) => {
+    return (
+      <OSUpdateModal
```
dkarski, line 45: 👍

packages/app/src/update/constants/environment.constant.ts
```diff
@@ -3,8 +3,8 @@
  * For licensing, see https://github.com/mudita/mudita-center/blob/master/LICENSE.md
  */
 
-export enum ReleaseSpace {
-  ProductionReleases = "latest",
-  PreReleases = "rc-latest",
-  DailyReleases = "daily-latest",
+export enum Environment {
```
dkarski, line 6: I'm wondering if to mark explicitly that this Environment is about OS, not the Center app. What do you think? 🤔

dkarski (approved):
LGTM ✅

packages/app/src/update/constants/environment.constant.ts
wojcik92michal: Good idea, definitely makes sense. Will change that

wojcik92michal: Changed in https://github.com/mudita/mudita-center/pull/1057/commits/1a969057b644e80909f6a8bc8e3d66c4bc287e8f

(not only `Environment`, but some other enums/interfaces)

dkarski (approved):

PR #1101, opened 2023-01-25
[CP-1782] [Passcode] No reaction after entering wrong passcode

dkarski:
Jira: [CP-1782]

**Description**


packages/app/src/device/observers/device-lock-time.observer.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    this.eventEmitter.on(DeviceServiceEvent.DeviceDisconnected, () => {
+      this.invoked = false
+      this.disconnected = true
+      this.unlocked = false
+    })
+
+    this.eventEmitter.on(DeviceServiceEvent.DeviceUnlocked, () => {
+      this.invoked = false
+      this.unlocked = true
```
wojcik92michal, line 53: Shouldn't it be added as well `this.disconnected = false`?

wojcik92michal (approved):

packages/app/src/device/observers/device-lock-time.observer.ts
dkarski: only when `DeviceDisconnected` is emitted 😉

lkowalczyk87 (approved):

PR #1112, opened 2023-02-06
[CP -1714] Search different behaviour for result display 

lkowalczyk87:
Jira: [CP-1714]

split search results from search preview for contacts and messages view

<details>
// put images here
</details>

[CP-1714]: https://appnroll.atlassian.net/browse/CP-1714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/app/src/search/actions/search.action.ts
+++ b/packages/app/src/search/actions/search.action.ts
@@ -8,10 +8,16 @@ import { SearchEvent } from "App/search/constants"
 import { SearchParams, SearchResult } from "App/search/dto"
 import { searchRequest } from "App/search/requests"
 
-export const search = createAsyncThunk<SearchResult | undefined, SearchParams>(
-  SearchEvent.SearchData,
-  async (payload) => {
-    const result = await searchRequest(payload)
-    return result.ok ? result.data : undefined
-  }
+const searchActionGenerator = (searchAction: SearchEvent) =>
+  createAsyncThunk<SearchResult | undefined, SearchParams>(
+    searchAction,
+    async (payload) => {
+      const result = await searchRequest(payload)
+      return result.ok ? result.data : undefined
+    }
+  )
+
+export const search = searchActionGenerator(SearchEvent.SearchData)
+export const searchPreview = searchActionGenerator(
+  SearchEvent.SearchDataPreview
 )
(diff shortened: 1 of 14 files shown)


dkarski (commented):
Almost works as expected... the empty search phrase triggering a re-rendering a result list

https://user-images.githubusercontent.com/17147149/217002530-c4cd0136-9123-4f59-ac2f-0441e1e9bae4.mov

packages/app/src/search/actions/search.action.ts
```diff
@@ -8,10 +8,16 @@ import { SearchEvent } from "App/search/constants"
 import { SearchParams, SearchResult } from "App/search/dto"
 import { searchRequest } from "App/search/requests"
 
-export const search = createAsyncThunk<SearchResult | undefined, SearchParams>(
-  SearchEvent.SearchData,
-  async (payload) => {
-    const result = await searchRequest(payload)
-    return result.ok ? result.data : undefined
-  }
+const searchActionGenerator = (searchAction: SearchEvent) =>
```
dkarski, line 11: 🟢 
A nice abstract to save a boilerplate. However, I'm not sure that is a good use case according to a DRY principle. 

> This is all well and good until you realize that now you’ve taken something simple and turned it into something really complicated that even you have a hard time understanding anymore.

source: [Balancing DRY and readability](http://jonkruger.com/blog/2014/02/24/balancing-dry-and-readability/) or something other [dry vs readability](https://www.google.com/search?q=dry+vs+readability&oq=DRY+vs+read&aqs=chrome.1.69i57j0i22i30l2j0i390l5.9191j0j7&sourceid=chrome&ie=UTF-8)

In general for this project... we don't create action generators to avoid mismatch and allow for easy maintenance of action logic when maintenance requires splitting or making it in another way 😉 However, I am fully open to your perspective if you have a different opinion 😇 🍀

lkowalczyk87: I totally agree that readability is more important than DRY. However in this case I have a feeling that action generator improves readability. Here's why:
- only actions are exported so abstraction have no impact readability in other files
- generator itself is really short and not too hard to understand, it basically passing one parameter (in other words it's not decreasing readability too much)
- and the most important point. it gives clear information that each of search actions do exactly the same, and only thing that differs those actions is name (which improves readability)

Regarding maintenance. In case of split action logic I agree that there is additional work to do, but if in future we want to move filtering/searching logic from other components to actions we can use same generator.

dkarski: Sure! I'm not going to convince you. I understand that we may have different opinions on this matter and I respect your perspective 🙇

dkarski (approved):

PR #1119, opened 2023-02-14
[CP-1787] add memory check before update

lkowalczyk87:
Jira: [CP-1787]

Add available memory check before the update process begins 

<details>

image

</details>


packages/app/src/__deprecated__/renderer/components/core/modal/modal.component.tsx
```diff
@@ -50,7 +50,7 @@ export const ButtonContainer = styled.div<{ buttonsPosition: ModalSize }>`
   ${({ buttonsPosition }) => getButtonsPosition(buttonsPosition)};
   display: flex;
   box-sizing: border-box;
-  padding-top: 4rem;
+  padding-top: 2.4rem;
```
dkarski, line 53: Are you sure that it is needed? I'm afraid that this change can be shared with other modals that use this core.

packages/app/src/__deprecated__/renderer/components/core/modal/modal.styled.elements.tsx
```diff
@@ -107,8 +107,6 @@ export const ModalSubTitle = styled(Text)`
 `
 
 export const Close = styled(Button)`
-  margin-top: -0.6rem;
-  margin-right: -0.8rem;
```
dkarski, line 111: The same case like above 🙄

packages/app/src/device-info/services/device-info.service.ts
```diff
@@ -30,4 +33,37 @@ export class DeviceInfoService {
...
+      >({
+        endpoint: Endpoint.DeviceInfo,
+        method: Method.Get,
+      })
+
+      if (!response.ok) {
+        return response
+      }
+
+      const freeSpaceInNoTSupportedDevice = +(
```
dkarski, line 50: 🟡 
Please use Number() 🙏  The casting (conversion of one data type to another) is making explicitly for this project.

packages/app/src/device-info/services/device-info.service.ts
```diff
@@ -30,4 +33,37 @@ export class DeviceInfoService {
...
+      const response = await this.deviceManager.device.request<
+        DeviceInfoRaw | NotSupportedDeviceInfo
+      >({
+        endpoint: Endpoint.DeviceInfo,
+        method: Method.Get,
+      })
+
+      if (!response.ok) {
+        return response
+      }
```
dkarski, line 48: 🟢 
What do you think to move this part of the code to a private method (example of name `getDeviceInfoRequest`) and sharing it between `getDeviceFreeSpace` and `getDeviceInfo` ? 🤔

packages/app/src/ui/components/modal-dialog/get-modal-dialog-style.helper.ts
```diff
@@ -30,7 +30,7 @@ const getModalSize = (size: ModalProps["size"]) => {
     case ModalSize.Small:
       return {
         width: "38rem",
-        padding: "2.4rem 2.4rem 4rem 2.4rem",
+        padding: "2.4rem 2.4rem 2.4rem 2.4rem",
```
dkarski, line 33: The same case like above 🙄 Im not sure that we should update a core/ui styles during on this issue...

dkarski (approved):

PR #1140, opened 2023-03-07
[CP-1778] Missed trackers of the updating process added to the project

dkarski:
Jira: [CP-1778]

**Description**

- Missed trackers of the updating process added to the project

[CP-1778]: https://appnroll.atlassian.net/browse/CP-1778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/app/src/update/actions/start-update-os/start-update-os.action.ts
+++ b/packages/app/src/update/actions/start-update-os/start-update-os.action.ts
@@ -6,7 +6,7 @@
 import { createAsyncThunk } from "@reduxjs/toolkit"
 import { AppError } from "App/core/errors"
 import { removeFile } from "App/device-file-system"
-import { DeviceType, DiagnosticsFilePath } from "App/device/constants"
+import { DiagnosticsFilePath } from "App/device/constants"
 import { setStateForInstalledRelease } from "App/update/actions/base.action"
 import {
   ReleaseProcessState,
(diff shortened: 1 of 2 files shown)


lkowalczyk87 (dismissed):

packages/app/src/update/actions/start-update-os/start-update-os.action.ts
```diff
@@ -62,6 +67,17 @@ export const startUpdateOs = createAsyncThunk<
 
     for (const release of releases) {
       state = getState() as RootState & ReduxRootState
+      const trackOsUpdateOptions: Omit<TrackOsUpdateOptions, "state"> = {
+        deviceType: state.device.deviceType as DeviceType,
```
patryk-sierzega, line 71: state.device.deviceType might be DeviceType or null. But what if it will be null and you'll cast it to DeviceType? It will be still null. Maybe it could be figured out differently? Like state.device.deviceType ?? DeviceType.MuditaPure?

packages/app/src/update/actions/start-update-os/start-update-os.action.ts
```diff
@@ -62,6 +67,17 @@ export const startUpdateOs = createAsyncThunk<
 
     for (const release of releases) {
       state = getState() as RootState & ReduxRootState
```
patryk-sierzega, line 69: About this state, wondering if we could deal with that type of getState once and not cast it every time we use it?

patryk-sierzega (approved):
LGTM, left some comments :)

I think we could think about refactoring function startUpdateOs, to me it seems too long and has to many responsibilities :)

packages/app/src/update/actions/start-update-os/start-update-os.action.ts
dkarski: The process is asynchronous so it is a reason, but I think that we could refactor it when time allows us to do.

PR #1142, opened 2023-03-07
[CP-1661] [Files manager] deleting status popup changes it's value

patryk-sierzega:
Jira: [CP-1661]

[CP-1661]: https://appnroll.atlassian.net/browse/CP-1661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dkarski:
@patryk-sierzega 👋 

I replaced your description...

**Before**

Zrzut ekranu 2023-03-7 o 10 06 30

**After**

Zrzut ekranu 2023-03-7 o 10 07 04

packages/app/src/files-manager/components/files-manager/files-manager.component.tsx
```diff
@@ -63,6 +66,7 @@ const FilesManager: FunctionComponent<FilesManagerProps> = ({
     uploadingInfo: false,
     uploadingFailed: false,
   })
+  const dispatch = useDispatch()
```
dkarski, line 69: 🟢 
We do not use the `useDispatch` hook in this project to no don't lose props control. 

However, it doesn't mean that we don't can start to use it. What do you think about discussing it at the front-end tech meeting?

dkarski (dismissed):
I left only one comment. However, it should not block this PR ✅ 

Congratulations on solving the first bug 👏

packages/app/src/files-manager/components/files-manager/files-manager.component.tsx
patryk-sierzega: Sure we can talk about it, used it because it was the easiest way for me :) I'll take a while a think about it :)

packages/app/src/files-manager/components/files-manager/files-manager.component.tsx
```diff
@@ -26,6 +27,7 @@ import { useLoadingState } from "App/ui"
 import { UploadFilesModals } from "App/files-manager/components/upload-files-modals/upload-files-modals.component"
 import { useFilesFilter } from "App/files-manager/helpers/use-files-filter.hook"
 import { getSpaces } from "App/files-manager/components/files-manager/get-spaces.helper"
+import { setDeletingFileLength } from "../../actions/base.action"
```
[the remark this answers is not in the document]
patryk-sierzega: That is interesting approach :D always used relative path... hm

lkowalczyk87 (dismissed):

dkarski (approved):
LGTM ✅ You are so precise 💪

lkowalczyk87 (approved):

PR #1155, opened 2023-03-16
E2E message mark unread/read tests

rajmundex:
Jira: [CP-XXX]

**Description**
Simple test to send a message and then to mark the conversation as unread and then read and unread again

<details>
image
</details>


packages/e2e/src/page-objects/messages.page.ts
```diff
@@ -129,6 +129,34 @@ class MessagesPage extends Page {
   > {
     return this.threadDropdownMarkAsReadButton.$("p")
   }
+  /** Returns list of last message text displayed on the conversation list and true/false depending on message unread/read status*/
+  async getLastMessages() {
+    const allLastMessages = await $$('//*[@data-testid="thread-last-message"]')
+
+    const messagesContent: string[] = await browser.executeAsync((done) => {
```
dkarski, line 136: `messagesContent` or `messagesContents`? 🤔 This is a list of the `messagesContent`?

`pseudoElements` it the same as `messagesContent`? Maybe it should be the same variable name to avoid to many new variable name contexts.

dkarski (dismissed):
PR is ok. I don't block it. I have only one comment on the `messagesContent` variable

packages/e2e/src/helpers/messages.helper.ts
```diff
@@ -1,5 +1,6 @@
 import MessagesPage from "../page-objects/messages.page"
 import MessagesConversationPage from "../page-objects/messages-conversation.page"
+import ModalMessagesPage from "../page-objects/modal-messages.page"
 
 /** Send message using send button */
 export const sendMessage = async (
```
lkowalczyk87, line 6: More a question than a suggestion. Why did you decide to move sendMessage and deleteConversationOnThreadList functions from MessagePage to helper?  I asking because those actions are available only on Message Page.

lkowalczyk87 (dismissed):

packages/e2e/src/helpers/messages.helper.ts
rajmundex: It is because pages focus on a part of app UI while helper can use selectors from few different pages (for example Delete message involves also confirmation available on separate modal page

packages/e2e/src/page-objects/messages.page.ts
rajmundex: cff6d8e82fa21faa9d1e36635830277c443a0981

dkarski (approved):

lkowalczyk87 (approved):

PR #1299, opened 2023-07-03
[CP-1964] [Update] Force update MC vs force update MOS

patryk-sierzega:
Jira: [CP-1964]

**Description**
- i think solution is a bit lazy, but works :D
- better solution would be to move os force update modal to modal manager, but I would advise to create task for it ;)
- there might be another bug in scenario when center is offline and center update is required cause there is modal about being offline and under it modal with updating center and we can do the last thing - this should be confirmed by tests and solved in different issue

edited (05.07.2023):
- i've removed previous solution and instead of it I've added larger zIndex for center force-update. This way center force-update has priority over os force-update and offline mode error :) And this eliminates third point of above list

[CP-1964]: https://appnroll.atlassian.net/browse/CP-1964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dkarski (approved):
Moving all models, including the OS force update modal, to the modal manager is okay. Please create a task for it. Thanks!

OskarMichalkiewicz (approved):

lkowalczyk87 (approved):

dkarski (approved):

lkowalczyk87 (approved):

OskarMichalkiewicz (approved):

PR #1300, opened 2023-07-03
[CP-2039] I want to be able to read and accept the Privacy policy the first time the application is launched after it has been updated

lkowalczyk87:
Jira: [CP-2039]

**Description**

<details>

image

</details>

[CP-2039]: https://appnroll.atlassian.net/browse/CP-2039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


out.txt
```diff
@@ -0,0 +1,139 @@
+
```
OskarMichalkiewicz, line 1: I guess this should be deleted 😁

packages/app/src/modals-manager/components/modals-manager.component.tsx
```diff
@@ -28,6 +31,14 @@ const ModalsManager: FunctionComponent<Props> = ({
...
+  const { privacyPolicyAccepted } = useSelector(
+    (state: ReduxRootState) => state.settings
+  )
+
+  if (privacyPolicyAccepted === false) {
+    return <PrivacyPolicyModal />
+  }
+
   return (
     <>
```
OskarMichalkiewicz, line 43:
```suggestion
    <>
        {!privacyPolicyAccepted && <PrivacyPolicyModal />}
```

out.txt
lkowalczyk87: deleted ;)

OskarMichalkiewicz (dismissed):

OskarMichalkiewicz (approved):

packages/app/src/settings/components/privacy-policy-modal/privacy-policy-modal.component.tsx
```diff
@@ -0,0 +1,114 @@
...
+  }, [collectingData, dispatch])
+
+  if (collectingData === undefined) {
+    return null
+  }
+
+  return (
+    <ModalDialog
+      open
+      zIndex={200}
```
patryk-sierzega, line 84: I know this is fine, but in the future we should avoid doing, cause every modal has it's own setting of zIndex. It should be managed by modal manager imho :)

packages/app/src/settings/services/settings.service.ts
```diff
@@ -32,7 +32,11 @@ export class SettingsService {
   }
 
   updateSettings({ key, value }: SettingsUpdateOption): SettingsValue {
-    this.store.set(key, value)
+    if (value !== undefined) {
+      this.store.set(key, value)
+    } else {
+      this.store.delete(key)
+    }
```
patryk-sierzega, line 39: Maybe update should only update and another function/method should be delivered to do deleting?

packages/app/src/settings/actions/delete-collecting-data.action.ts
```diff
@@ -0,0 +1,17 @@
...
+import { createAsyncThunk } from "@reduxjs/toolkit"
+import { SettingsEvent } from "App/settings/constants"
+import { updateSettings } from "App/settings/requests"
+
+export const deleteCollectingData = createAsyncThunk(
+  SettingsEvent.DeleteCollectingData,
+  async () => {
+    await updateSettings({ key: "collectingData", value: undefined })
+
+    return
```
patryk-sierzega, line 15: Do you need this return without value? I think js/ts will return undefined when there is no return :)

packages/app/src/modals-manager/components/modals-manager.component.tsx
```diff
@@ -28,6 +31,14 @@ const ModalsManager: FunctionComponent<Props> = ({
   deviceInitializationFailedModalShowEnabled,
   hideModals,
 }) => {
+  const { privacyPolicyAccepted } = useSelector(
+    (state: ReduxRootState) => state.settings
+  )
+
+  if (privacyPolicyAccepted === false) {
+    return <PrivacyPolicyModal />
+  }
```
patryk-sierzega, line 40: a bit nit-picking ;D i've got mixed feelings about value === false or value === true, maybe '!privacyPolicyAccepted' would be better? What do you think?

patryk-sierzega (approved):

PR #1322, opened 2023-07-21
[CP-2091]You can do both Force Center Update and start Force OS Update because modal buttons are not locked when Force Center Update starts.

OskarMichalkiewicz:
Jira: [CP-2091]

**Description**

<details>
// put images here
</details>

[CP-2091]: https://appnroll.atlassian.net/browse/CP-2091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/app/src/modals-manager/constants/modal-layers.enum.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Copyright (c) Mudita sp. z o.o. All rights reserved.
+ * For licensing, see https://github.com/mudita/mudita-center/blob/master/LICENSE.md
+ */
+
+export enum ModalLayers {
```
lkowalczyk87, line 6: 🟡  I think that layers names should represent the system domain/feature or even part domain/feature eg. Default = 5, PrivacyPolicy, MCForceUpdace

packages/app/src/ui/components/modal-dialog/modal-dialog.component.tsx
```diff
@@ -86,7 +87,7 @@ export const ModalDialog: FunctionComponent<ModalDialogProps> = withTheme(
         isOpen={open}
         // AUTO DISABLED - fix me if you like :)
         // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-        style={getModalDialogStyle({ zIndex, size, theme })}
+        style={getModalDialogStyle({ zIndex: layer, size, theme })}
```
patryk-sierzega, line 90: so, basically, there will be still zIndex underneath every modal, but from now on we call it layer? :D

patryk-sierzega (approved):

packages/app/src/ui/components/modal-dialog/modal-dialog.component.tsx
OskarMichalkiewicz: pretty much, yeah. I remember that idea to call it layer or something similar originated during my conversation with @lkowalczyk87 but ofc this is not mandatory. I can change the name to zIndex or smth, if majority will agree to diffirent a name.

lkowalczyk87 (approved):

PR #1338, opened 2023-08-02
[CP-2055][Help] in offline mode incorrect topics are displayed

OskarMichalkiewicz:
Jira: [CP-2055]

** download default help items on setup **

<details>
// put images here
</details>

[CP-2055]: https://appnroll.atlassian.net/browse/CP-2055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dkarski (approved):

packages/app/scripts/downloadHelpItems.ts
```diff
@@ -0,0 +1,29 @@
...
+})
+import { normalizeHelpData } from "../../app/src/__deprecated__/renderer/utils/contentful/normalize-help-data"
+;(async () => {
+  try {
+    await fs.ensureDir(path.resolve(path.join("src", "main")))
+    const jsonPath = path.join("src", "help", "default-help.json")
+
+    const url = `${process.env.MUDITA_CENTER_SERVER_URL}/help`
+    const { data } = await axios.get(url, {
+      params: { limit: 6 },
```
lkowalczyk87, line 20: Why 6? If that's because the server uses 6 as the default value for the limit variable maybe we can leave that decision to endpoint logic.

OskarMichalkiewicz: I basically used news download script and modified it. That is something I've missed. It may also mean that the news script should not use limit param either🤔.

OskarMichalkiewicz: I've removed the limit param as I didn't see any changes to the result with or without it

lkowalczyk87 (approved):

dkarski (approved):

PR #1362, opened 2023-08-21
[CP-1873] Add the Phone Number Endpoint to make easier communication between OS <-> Center

patryk-sierzega:
Jira: [CP-1873]

**Description**

**Questions**
- should we add new service for phone numbers and inject it into ThreadService and ContactService or keep request in src/common/requests/phone-number.request.ts?
- I've renamed types PureContact, PureMessage, PureThread adding prefix 'Pure', so we won't have confusion which type is correct, maybe we should also change other types which are used when getting response from device? Cause naming two types Thread and then using those in one file and renaming one of them during import seemed to me as bad approach
- I think we should consider checking all tests and question their necessity, it could be done before refactoring connectors to hooks, it might just ease that refactor process, what do you think?
- names of variables in types connected to Pure and serial port have names with 'ID' like numbersIDs, threadID, numberID, contactID. Those names comes from device and changing them to version with 'Id' is a little bit overkill and can cause confusion. I would keep those names.

[CP-1873]: https://appnroll.atlassian.net/browse/CP-1873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dkarski (commented):
🟡 

these changes do not include interface synchronization for actions such as
- adding a new contact
- editing a new contact
- the appearance of a new contact on the Pure device
- edit new contact on Pure device

I would also make sure that changes to the thread are taken into account at
- creating a new thread in the Mudita Center application
- the appearance of a new thread/message on the Pure device

packages/app/src/contacts/presenters/contact.presenter.test.ts
```diff
@@ -14,6 +14,7 @@ const pureContact: PureContact = {
   blocked: false,
   favourite: true,
   numbers: ["500400300"],
+  numbersIDs: ["1"],
```
dkarski, line 17: Please keep name conventions with the current project or rename it for all 😉 
```suggestion
  numberIds: ["1"],
```

The current state of the project
```
id -> // bad ID
ids -> // bad IDs
numberIds  // bad -> numbersIDs / numberIDs
numberId  // bad -> numberID

```

dkarski:
> should those fields 'numberIDs' be optional or mandatory in types/interfaces?

In my opinion, should be mandatory property. The force os update allows us to keep the latest interface obligatory. However, even if we would like to maintain compatibility (with old Pure models), we can return empty values at the presenter layer.

packages/app/src/messages/services/thread.service.ts
```diff
@@ -153,34 +179,6 @@ export class ThreadService {
     )
   }
 
-  // AUTO DISABLED - fix me if you like :)
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore
-  // the method is commented until os part will be implemented as CP-1232
-  private async getThreadRequest(id: string): Promise<RequestResponse<Thread>> {
```
dkarski, line 160: I see that you handled the remarks  👍 Good to know you're looking in there 😉 😅

packages/app/src/contacts/reducers/contacts.interface.ts
```diff
@@ -14,6 +14,8 @@ export interface Contact {
   lastName?: string
   primaryPhoneNumber?: string
   secondaryPhoneNumber?: string
+  numbers?: string[]
+  numbersIDs?: string[]
```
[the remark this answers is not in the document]
patryk-sierzega: Yeap, i forgot about it, focused too much on making it working ;d

packages/app/src/contacts/presenters/contact.presenter.test.ts
patryk-sierzega: packages/app/src/device/types/mudita-os/serialport-request.type.ts

here are interfaces PureMessage, PureThread and PureContact - we probably are keeping this convention 'ID' because of Pure, we can change it, sure, but in my opinion it would add confusion to the code like:
- we receive data as 'ID', then we rename it to 'Id', next we use it

and i agree that name 'numbersIDs' might not be the best :D

packages/app/src/contacts/services/contact.service.ts
```diff
@@ -35,10 +37,23 @@ export class ContactService {
         },
       })
 
+    const contact = response.data as PureContact
+    const phoneNumberIds = contact?.numbersIDs ?? []
+
+    const phoneNumberPromises = phoneNumberIds.map(() => {
+      return getPhoneNumberRequest(this.deviceManager, phoneNumberIds[0])
```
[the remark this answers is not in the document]
patryk-sierzega: We already talked about this, it seems like art for art's sake. We could, but I'll ask question about it for the rest of frontend team when final PR will arrive.

packages/app/src/common/requests/phone-number.request.ts
```diff
@@ -0,0 +1,26 @@
...
+): Promise<PhoneNumberResponse> => {
+  const phoneNumberResponse =
+    await deviceManager.device.request<GetPhoneNumberResponseBody>({
+      endpoint: Endpoint.PhoneNumber,
+      method: Method.Get,
+      body: {
+        numberID: Number(phoneNumberId),
+      },
+    })
+  return phoneNumberResponse.data as PhoneNumberResponse
```
dkarski, line 25: can we do it without association? 🤔

packages/app/src/contacts/reducers/contacts.interface.ts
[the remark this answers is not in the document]
dkarski: What with that? Does your change affect the returned interface by any chance?

packages/app/src/contacts/services/contact.service.ts
```diff
@@ -26,46 +26,37 @@ export class ContactService {
...
-        data: ContactPresenter.mapToContact(response.data),
-      }
-    } else {
-      return {
-        status: RequestResponseStatus.Error,
-        error: { message: "Get contact: Something went wrong" },
-      }
-    }
-  }
+      const phoneNumbersResponses = await getPhoneNumbersRequest(
```
OskarMichalkiewicz, line 40: why plural?

OskarMichalkiewicz (approved):

dkarski (approved):

OskarMichalkiewicz (approved):

OskarMichalkiewicz (approved):

dkarski (approved):

PR #1440, opened 2023-10-12
[CP-2203] OS update badge status enhancements

OskarMichalkiewicz:
JIRA Reference: [CP-XXX]

### :memo: Description ️

To enhance update translation of both normal and update statuses into one I've restructured the hook and used redux hooks instead of prop drilling.

- [x] redux selectors are used in components / prop drilling is reduce

- [ ] change went through the QA process
- [ ] translations are updated in dedicated application
- [ ] [CHANGELOG.md](./CHANGELOG.md) is updated


packages/app/src/overview/components/overview-screens/pure-overview/overview-content.component.tsx
```diff
@@ -22,11 +22,6 @@ interface OverviewContentProps {
...
   readonly networkName: string
   readonly networkLevel: number | undefined
-  readonly pureOsAvailable: boolean
-  readonly pureOsDownloaded: boolean
-  readonly checkForUpdatePerformed: boolean
-  readonly checkForUpdateInProgress: boolean
-  readonly checkForUpdateFailed: boolean
   readonly osVersion: string
   readonly caseColour: CaseColor | undefined
   readonly lastBackupDate: Date | undefined
```
patryk-sierzega, line 27: I know this is not your code, but why do we need those readonly here and we don't use those anywhere else? :D Maybe we could throw them out? What do you think?

packages/app/src/overview/components/overview-screens/helpers/use-update-flow-state.hook.tsx
```diff
@@ -3,84 +3,125 @@
...
+    silentCheckForUpdate,
+    checkForUpdateState,
+    needsForceUpdate,
+    allDownloaded,
+    availableReleasesForUpdate,
+    error,
+  ])
   return {
     checkForUpdateLocalState,
   }
```
patryk-sierzega, line 127: is it necessary to return object with one field? :)

packages/app/src/overview/components/overview-screens/helpers/use-update-flow-state.hook.tsx
```diff
@@ -3,84 +3,125 @@
...
+      setCheckForUpdateLocalState(CheckForUpdateLocalState.Failed)
     } else if (
-      silentCheckForUpdateState === SilentCheckForUpdateState.Loaded ||
-      checkingForUpdateState === CheckForUpdateState.Loaded ||
-      checkingForUpdateState === CheckForUpdateState.Performed
+      [CheckForUpdateState.Loaded, CheckForUpdateState.Performed].includes(
+        checkForUpdateState
+      ) ||
+      SilentCheckForUpdateState.Loaded === silentCheckForUpdate
     ) {
```
patryk-sierzega, line 114: Under this else-if the is no 'else', I am wondering that it might be a case when none of above condition is true, maybe we should set something when none of this cases is true? Or maybe not and leave it undefined...

patryk-sierzega (dismissed):

packages/app/src/overview/components/overview-screens/helpers/use-update-flow-state.hook.tsx
OskarMichalkiewicz: I was thinking that maybe a good idea would be to return a set of booleans representing those statuses. What do you think?

packages/app/src/overview/components/overview-screens/helpers/use-update-flow-state.hook.tsx
OskarMichalkiewicz: I think that a lot of issues with those badges were caused by the case where none of the conditions were true and then we defaulted to some text that did not match the situation. I think returning loading status or null would be nice (so that nothing will render based on those)

packages/app/src/overview/components/overview-screens/pure-overview/overview-content.component.tsx
OskarMichalkiewicz: I think we can get rid of those :D

patryk-sierzega (approved):

PR #1558, opened 2023-11-13
[CP-1977] [SOSU] Update with more than one devices connected to PC

OskarMichalkiewicz:
JIRA Reference: [CP-1977]

### :memo: Description ️

-

[CP-1977]: https://appnroll.atlassian.net/browse/CP-1977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/app/src/device-manager/services/device-manager.service.ts
```diff
@@ -25,6 +25,8 @@ export class DeviceManager {
   public currentDevice: Device | undefined
   public devicesMap = new Map<string, Device>()
   public currentDeviceInitializationFailed = false
+  private lastSerialNumber = ""
+  private lastProductId = ""
```
dkarski, line 29:
```suggestion
  private lastDeviceId = "" // as `${lastSerialNumber}-${lastProductId}`
```

I suggest considering the use of a single field `lastDeviceId` instead of two (`lastSerialNumber` and `lastProductId`)

Why it's important in my opinion:

1. **Code Clarity:** One field with a meaningful name can be more readable than two separate variables, especially when both values are related.
2. **Easier Management:** Grouping data into a single structure facilitates its management and reduces the risk of errors when manipulating these values.
3. **Conciseness:** Using template strings enables an elegant and concise way of combining values, contributing to clearer code.

dkarski (commented):
- The changes are fine ✅ 
- It would be nice if you could consider my comments, or at least refer to them. We can catch up on a huddle if you'd like. 💚 💚 💚 
- Please fix the "check." 🔴

packages/app/src/device-manager/services/device-manager.service.ts
```diff
@@ -66,6 +68,13 @@ export class DeviceManager {
     if (this.currentDevice) {
       return
     }
+    if (this.updating) {
+      if (!+this.lastSerialNumber && port.productId !== this.lastProductId) {
```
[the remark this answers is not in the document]
patryk-sierzega: Agreed, avoid using this type of conversion :)

packages/app/src/device-manager/services/device-manager.service.ts
```diff
@@ -90,26 +101,37 @@ export class DeviceManager {
...
-  public removeDevice(path: string): void {
+  public removeDevice(path: string, productId: string): void {
     this.devicesMap.delete(path)
 
     if (this.currentDevice?.path === path) {
       if (this.devicesMap.size > 0) {
         this.currentDevice = this.devicesMap.values().next().value as Device
+        this.lastProductId = this.updating ? this.lastProductId : productId
+        this.lastSerialNumber = this.updating
+          ? !+this.lastSerialNumber
```
patryk-sierzega, line 112: here to, please change that '+' into Number()

packages/app/src/device-manager/services/device-manager.service.ts
```diff
@@ -122,6 +144,12 @@ export class DeviceManager {
     }
 
     this.currentDevice = newCurrentDevice
+    this.lastProductId = this.updating ? this.lastProductId : productId
+    this.lastSerialNumber = this.updating
+      ? !+this.lastSerialNumber
```
patryk-sierzega, line 149: '+' to Number()

packages/app/src/update/services/device-update.service.ts
```diff
@@ -53,6 +54,19 @@ export class DeviceUpdateService {
       )
     }
     this.beforeUpdateOsVersion = deviceInfoResult.data.osVersion
+    if (
+      isVersionGreaterOrEqual(
+        this.beforeUpdateOsVersion,
+        payload.fileName.split("-")[1]
```
patryk-sierzega, line 60: suggest to store it in variable 'payload.fileName.split("-")[1]' and then pass to function for clarity :)

packages/app/src/device-manager/services/device-manager.service.ts
patryk-sierzega: If necessary, i would suggest to use complex type like lastIds: {productId: string, serialNumber: string}, because in code those fields are used separately. If stored as template string we would need to extract single value from string.
For me those two fields are fine :)

patryk-sierzega (approved):

dkarski (approved):

dkarski:
@OskarMichalkiewicz to draft?

PR #1619, opened 2023-12-27
[CP-2311] Moved API PoCs

mkurczewski:
JIRA Reference: [CP-2311]

### :memo: Description ️

-

[CP-2311]: https://appnroll.atlassian.net/browse/CP-2311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lkowalczyk87 (approved):

libs/generic-view/feature/src/lib/recursive-layout/recursive-layout.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * For licensing, see https://github.com/mudita/mudita-center/blob/master/LICENSE.md
+ */
+
+import React, { FunctionComponent } from "react"
+import { useSelector } from "react-redux"
+import { ReduxRootState } from "Core/__deprecated__/renderer/store"
+import apiComponents from "generic-view/ui"
+import { APIFC, withLayout } from "generic-view/utils"
+
+interface MetaData {
```
dkarski, line 12: 🟢 

```suggestion
interface Props {
```

Instead of the name `MetaData`, consider renaming the interface to `Props`, which more accurately reflects that the interface defines component properties.

I suggest this change to facilitate using this interface in other components and to avoid the need for creating special names.

libs/device/models/src/lib/device-communication-event.ts
```diff
@@ -0,0 +1,20 @@
...
+export const ApiSerialPortEvents = {
+  DataReceived: "api-serial-port-data-received",
+} as const
+
+export type ApiSerialPortEvent =
+  (typeof ApiSerialPortEvents)[keyof typeof ApiSerialPortEvents]
+
+export const ApiSerialPortToRendererEvents = {
+  Connected: "api-serial-port-connected",
+  Disconnected: "api-serial-port-disconnected",
```
dkarski, line 15: 🟡 
I suggest changing 
```suggestion
  Detached: "api-serial-port-detached",
```
because the term "Disconnected" can be misleading. In the case of an application allowing disconnection of the device, the device remains physically attached to the USB port, corresponding to the "Attached" state.

This approach more accurately reflects the state where the device is visible in the USB stack but has not yet initiated a connection, even when the user disconnects the device in the application layer.

libs/device/models/src/lib/device-communication-event.ts
```diff
@@ -0,0 +1,20 @@
...
+  InitializationFailed: "api-serial-port-initialization-failed",
```
dkarski, line 16: 🟡 
 suggest changing the state name
```suggestion
  ConnectionFailed: "api-serial-port-connection-failed",
```

This change more accurately reflects the nature of the error, which occurs during an attempt to establish a connection via the serial port interface rather than specifically during the initialization process. 

This modification should more appropriately describe the situation where an error occurs during an attempt to establish a connection with the serial port interface, not necessarily during the initialization process.

dkarski (changes requested):
Thanks for the PR! It looks great, but I have one note. After initializing Mudita Pure and Mudita Harmony, the navigation view for the elements seems off due to issues with API Device-related elements. It's a minor thing, but the changes can't be merged until it's fixed.

Zrzut ekranu 2024-01-4 o 05 54 49
Zrzut ekranu 2024-01-4 o 05 55 08

dkarski:
Please make sure to include the JIRA Reference in both the description and the title. It helps us keep everything organized. Thanks! 🙃

libs/generic-view/feature/src/lib/recursive-layout/recursive-layout.tsx
lkowalczyk87: In this case, I think `MetaData` is more meaningful than the generic name `Props`. So I would suggest sticking to `MetaData` or considering something more unique like `RecursiveLayoutProps`

lkowalczyk87:
> Thanks for the PR! It looks great, but I have one note. After initializing Mudita Pure and Mudita Harmony, the navigation view for the elements seems off due to issues with API Device-related elements. It's a minor thing, but the changes can't be merged until it's fixed.
> 
> Zrzut ekranu 2024-01-4 o 05 54 49 Zrzut ekranu 2024-01-4 o 05 55 08
@dkarski  are you sure you pulled all changes before reviewing?
image

dkarski:
> > Thanks for the PR! It looks great, but I have one note. After initializing Mudita Pure and Mudita Harmony, the navigation view for the elements seems off due to issues with API Device-related elements. It's a minor thing, but the changes can't be merged until it's fixed.
> > Zrzut ekranu 2024-01-4 o 05 54 49 Zrzut ekranu 2024-01-4 o 05 55 08
> > @dkarski  are you sure you pulled all changes before reviewing?
> > image

I'm very sorry! I didn't have the latest changes loaded. Sorry, I wasn't aware that I was on this branch before

dkarski (dismissed):
Due to my mistake, I am withdrawing the request for changes by approving

libs/device/models/src/lib/device-communication-event.ts
lkowalczyk87: even more correct name would be `api-serial-port-closed` because the event is emitted as a response for serial port `close` event  :)

libs/generic-view/feature/src/lib/recursive-layout/recursive-layout.tsx
dkarski: Thanks for your perspective. I believe the property doesn't need to have a unique name until it's exported, and the idea of `RecursiveLayoutProps` works for me. I've marked my opinion with a green dot, so the author can decide what to do with it.

mkurczewski: This interface is used only by this component itself (no `export`), so it's name doesn't really matter. Just "Properties" should be fine.

lkowalczyk87 (dismissed):

lkowalczyk87 (approved):

PR #1655, opened 2024-01-15
[CP-2395] Add overview service

lkowalczyk87:
JIRA Reference: [CP-2395]

### :memo: Description ️

-

[CP-2395]: https://appnroll.atlassian.net/browse/CP-2395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/device/models/src/lib/overview/overview-config.ts
```diff
@@ -0,0 +1,70 @@
...
+    showImg?: boolean //default true
+    imgVariant?: string
+    showSerialNumber?: boolean //default true
+    showAbout?: boolean //default false
+    aboutTitle?: string //default defined by translation
+    aboutSubtitle?: string
+    aboutIcon?: IconType
+    aboutFields?: Array<DetailListFieldConfig> //default []
+  }
+  sections?: Array<OverviewSectionsConfig> //default []
```
patryk-sierzega, line 69: if you use default values maybe it would be better to avoid `| undefined`?

patryk-sierzega (approved):

libs/device/models/src/lib/overview/overview-config.ts
lkowalczyk87: good point, it'll be split into two separate interfaces: one with possible undefined values (API response may override default settings or skip filed if default setting should be used) and one without undefined values

PR #1666, opened 2024-01-17
[CP-2165] I want to switch to another device

patryk-sierzega:
JIRA Reference: [CP-2165]

### :memo: Description ️

-

[CP-2165]: https://appnroll.atlassian.net/browse/CP-2165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dkarski:
🟡 
Sure, please pay attention to the sizes of elements and the areas of empty margins, etc., as there is a difference between design and implementation

Zrzut ekranu 2024-01-19 o 09 00 39
Zrzut ekranu 2024-01-19 o 09 00 46

left: feature branch | right: design

Zrzut ekranu 2024-01-19 o 08 58 22
Zrzut ekranu 2024-01-19 o 08 59 13

libs/core/__deprecated__/renderer/components/rest/menu/menu-bottom.component.tsx
```diff
@@ -51,14 +52,52 @@ interface Props {
...
+      />
+    </DeviceButtonWrapper>
+  )
+
+  const isChangeDevice =
+    !dataSyncInProgress &&
+    devices.length > 1 &&
+    history.location.pathname !== URL_MAIN.news
+
+  const ChangeDeviceButton = (
```
dkarski, line 85: 🟢 
I'm considering moving the `SelectDeviceButton` and `ChangeDeviceButton` components to separate files. This would allow for:

1. **Modularity:** Having each component in a separate file would facilitate code organization and development.
2. **Clearer Project Structure:** Separate files would make navigation and understanding of the component hierarchy in the application easier.

libs/core/__deprecated__/renderer/locales/default/en-US.json
```diff
@@ -197,7 +199,7 @@
   "component.updateProgressModalTitle": "Updating Mudita Center",
   "module.availableDeviceList.headerTitle": "Connected devices",
   "module.availableDeviceList.subheaderTitle": "Select the device you want to use",
-  "module.availableDeviceList.serialNumber": "Serial Number",
+  "module.availableDeviceList.serialNumber": "Serial number",
```
dkarski, line 202: thanks ❤️ x2

libs/core/__deprecated__/renderer/locales/default/en-US.json
```diff
@@ -74,6 +74,8 @@
   "component.devModeHeader": "DEV",
   "component.devModeToggle": "Toggle dev window",
   "component.devModeWrapperName": "Developer Options",
+  "component.deviceSelection.selectDevice": "Select Device",
```
dkarski, line 77: thanks ❤️

libs/core/device-manager/reducers/device-manager.reducer.ts
```diff
@@ -104,5 +107,17 @@ export const deviceManagerReducer = createReducer<DeviceManagerState>(
           activeDeviceId: undefined,
         }
       })
+      .addCase(openSelectDeviceDrawer, (state) => {
```
dkarski, line 110: ⚪ Neutral
maybe toggler? less code

```
toggleSelectDeviceDrawer(true)
toggleSelectDeviceDrawer(false)
```

libs/core/device-select/components/drawer-device.component.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { DeviceType } from "Core/device"
+import { CaseColor } from "Core/device/constants"
+import { getDeviceTypeName } from "Core/discovery-device/utils/get-device-type-name"
+import { intl } from "Core/__deprecated__/renderer/utils/intl"
+
+const Device = styled("div")<{ active: boolean }>`
+  display: flex;
+  justify-content: flex-start;
+  margin: 1rem 0.5rem 1rem 0.5rem;
+
```
dkarski, line 28: 🟡 

Your code was missing animations for the hover. Introduced the transition property for smoother animations. Now, changes in background color occur gradually and more smoothly when hovering.

```suggestion
  margin: 1rem 0.5rem 1rem 0.5rem;
  transition: background-color ${transitionTime("veryQuick")} ${transitionTimingFunction("smooth")};
```

libs/core/device-select/constants/event.constant.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright (c) Mudita sp. z o.o. All rights reserved.
+ * For licensing, see https://github.com/mudita/mudita-center/blob/master/LICENSE.md
+ */
+
+export enum DeviceSelectEvent {
+  OpenSelectDeviceDrawer = "device-select_open-select-device-drawer",
+  CloseSelectDeviceDrawer = "device-select_close-select-device-drawer",
+}
```
dkarski, line 9: ⚪ Neutral

What are your thoughts on using this algorithm for Redux actions?
```
DEVICE_SELECT-CLOSE_SELECT_DEVICE_DRAWER
```

DEVICE_SELECT - Event ID
CLOSE_SELECT_DEVICE_DRAWER? - Redux Action Name

dkarski (commented):
I have no objections to the code itself. I will approve after incorporating design-related corrections.

libs/core/device-select/constants/event.constant.ts
patryk-sierzega: I am fully convinced, I think I would prefer lowercase :)

libs/core/__deprecated__/renderer/components/rest/menu/menu-bottom.component.tsx
patryk-sierzega: Not sure about that, SelectDeviceButton and ChangeDeviceButton are only implementations of DeviceButton. I am just passing props to that component, nothing more.

I've moved it out of return to be more readable.

libs/core/device-manager/reducers/device-manager.reducer.ts
patryk-sierzega: I think u are suggesting method to simply update state, not a toggler. Toggles changes value to the opposite.

libs/core/device-select/components/drawer-device.component.tsx
patryk-sierzega: I am not missing that, there is no mentioning this in Figma

libs/core/device-manager/reducers/device-manager.reducer.ts
dkarski: Yes, you understood the essence of my suggestion. However, remember that you don't have to consider it at all, as the change is great.

dkarski (approved):
🟢 
I would suggest removing the cursor pointer change on the active state, as it feels more natural. Instead, let's add a hover animation for devices. This is a minor adjustment, and it's better to do it right away without a separate pull request and an additional ticket. Just need a confirmation from Renata on the Slack channel.

libs/core/device-select/components/drawer-device.component.tsx
dkarski: https://github.com/mudita/mudita-center/pull/1666#pullrequestreview-1836090683

libs/core/__deprecated__/renderer/components/rest/menu/menu-bottom.component.tsx
dkarski: Moving this outside the main component block could be a compromise for me. However, as I mentioned at the beginning of the comment, please don't get blocked by my remark, as it is not blocking at all.

dkarski: Moving this outside the main component block could be a compromise for me. However, as I mentioned at the beginning of the comment, please don't get blocked by my remark, as it is not blocking at all.

dkarski (dismissed):

libs/core/__deprecated__/renderer/components/rest/menu/menu-bottom.component.tsx
```diff
@@ -71,6 +116,8 @@ const MenuBottom: FunctionComponent<Props> = ({ dataSyncInProgress }) => {
           </Text>
         </SyncProgressWrapper>
       )}
+      {isSelectDevice && SelectDeviceButton}
+      {isChangeDevice && ChangeDeviceButton}
       {!deviceInitialized && devices.length > 0 && (
```
dkarski, line 121: please remove this part of the code if you would like to avoid situation like in the bellow screen 👇 
Zrzut ekranu 2024-01-30 o 08 55 33

dkarski (dismissed):
🟢 

I've reviewed your code, and the current method using polling is OK, but it might be worth considering the use of IPC in Electron for greater efficiency. IPC would allow for immediate response to changes without the need for constant state polling.

dkarski (dismissed):

dkarski (dismissed):

dkarski (approved):

PR #1676, opened 2024-01-23
Overview summary mapper

mkurczewski:
JIRA Reference: [CP-XXX]

### :memo: Description ️

-


libs/generic-view/feature/src/lib/generic-modals.tsx
```diff
@@ -13,9 +13,21 @@ interface Props {
 }
 
 export const GenericModals: FunctionComponent<Props> = ({ viewKey }) => {
-  const { views } = useSelector((state: ReduxRootState) => state.genericViews)
+  const devicesConfiguration = useSelector(
+    (state: ReduxRootState) => state.genericViews.devicesConfiguration
```
lkowalczyk87, line 17: please add selector

libs/generic-view/feature/src/lib/generic-view.tsx
```diff
@@ -18,13 +18,27 @@ export const GenericView: FunctionComponent = () => {
     subviewKey?: string
   }>()
   const currentViewKey = subviewKey || viewKey
-  const { views } = useSelector((state: ReduxRootState) => state.genericViews)
+  const devicesConfiguration = useSelector(
+    (state: ReduxRootState) => state.genericViews.devicesConfiguration
```
lkowalczyk87, line 22: consider refactoring by adding selectors

libs/generic-view/feature/src/lib/recursive-layout.tsx
```diff
@@ -18,10 +18,13 @@ export const RecursiveLayout: FunctionComponent<Properties> = (
   recursiveComponentMetadata
 ) => {
   const { viewKey, componentKey } = recursiveComponentMetadata
-  const { childrenKeys, component } = useSelector(
-    (state: ReduxRootState) =>
-      state.genericViews.views?.[viewKey]?.layout?.[componentKey]
-  )
+  const { childrenKeys, component } = useSelector((state: ReduxRootState) => {
+    const features =
```
lkowalczyk87, line 22: consider refactoring by adding selectors

libs/generic-view/utils/src/lib/with-config/with-config.tsx
```diff
@@ -7,15 +7,28 @@ import React, { ComponentType } from "react"
 import { useSelector } from "react-redux"
 import { ReduxRootState } from "Core/__deprecated__/renderer/store"
 import { RecursiveComponent } from "../models/api-fc.types"
+import { createSelector } from "reselect"
+
+type Keys = { viewKey: string; componentKey: string }
+
+const configSelector = createSelector(
+  (state: ReduxRootState) => state.genericViews.devicesConfiguration,
```
lkowalczyk87, line 15: you can pass selector as a parameter so if we ever change the field name it will be hidden behind "facade" and only the selector will require an update. same for other similar cases

libs/generic-view/views/src/lib/mc-overview/summary/summary.ts
```diff
@@ -0,0 +1,135 @@
...
+        [SummaryKeys.Image]: {
+          component: "image",
+          layout: {
+            flexPlacement: {
+              alignSelf: "center",
+            },
+            padding: "32px 27px 16px 27px",
+          },
+          config: {
+            // TODO: implement support for config.imgVariant
```
lkowalczyk87, line 36: 👍

libs/generic-view/utils/src/lib/with-config/with-config.tsx
mkurczewski: Utils cannot import from `generic-view/store` because there are circular dependencies.

PR #1687, opened 2024-01-30
[CP-2163] Fix for Initialization Abort Process and Enhancement of Discovery Mode Logic

dkarski:
JIRA Reference: [CP-2163]

#### Fix for Initialization Abort Process: 5bcb133a38f5b8be55da95848259959f1942caf6

- Enhanced the application's functionality so that when a user cancels the initialization process, the application correctly redirects to the news view.

#### Optimization of useDiscoveryRedirectEffect Logic: 15f01c3885d21736fa18ed806abf69573e37d695

- Improved the process of entering discovery mode, activated when a user manually checks the app version in settings.
- Fixed an issue with an endless loader that occurred when a user disconnects and reconnects the device.

#### Additional Fix: 15f01c3885d21736fa18ed806abf69573e37d695

- Resolved an issue with the 'disconnect' button, which did not function correctly in the previous version.

[CP-2163]: https://appnroll.atlassian.net/browse/CP-2163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/core/core/hooks/use-device-detached-effect.ts
+++ b/libs/core/core/hooks/use-device-detached-effect.ts
@@ -16,7 +16,7 @@ import { isActiveDeviceProcessingSelector } from "Core/device-manager/selectors/
 import { deactivateDevice } from "Core/device-manager/actions/deactivate-device.action"
 import { cancelOsDownload } from "Core/update/requests"
 import { URL_ONBOARDING } from "Core/__deprecated__/renderer/constants/urls"
-import { useRedirectOnActiveDeviceDetached } from "Core/overview/components/overview-screens/pure-overview/use-redirect-on-active-device-detached.hook"
+import { useDeactivateDeviceAndRedirect } from "Core/overview/components/overview-screens/pure-overview/use-deactivate-device-and-redirect.hook"
 import { getDevicesSelector } from "Core/device-manager/selectors/get-devices.selector"
 import { setSelectDeviceDrawerOpen } from "Core/device-select/actions/set-select-device-drawer-open.action"
 
@@ -29,7 +29,7 @@ export const useDeviceDetachedEffect = () => {
     ({ update }: ReduxRootState) => update.downloadState
   )
   const devices = useSelector(getDevicesSelector)
-  const redirectOnActiveDeviceDetached = useRedirectOnActiveDeviceDetached()
+  const deactivateDeviceAndRedirect = useDeactivateDeviceAndRedirect()
 
   useEffect(() => {
     return registerDeviceDetachedListener(
@@ -57,7 +57,7 @@ export const useDeviceDetachedEffect = () => {
           history.push(URL_ONBOARDING.welcome)
           return
         }
-        redirectOnActiveDeviceDetached()
+        deactivateDeviceAndRedirect()
       }
     )
   }, [
@@ -67,6 +67,6 @@ export const useDeviceDetachedEffect = () => {
     downloadProcessing,
     dispatch,
     history,
-    redirectOnActiveDeviceDetached,
+    deactivateDeviceAndRedirect,
   ])
 }
--- a/libs/core/core/hooks/use-device-detached-effect.ts
+++ b/libs/core/core/hooks/use-device-detached-effect.ts
@@ -18,7 +18,6 @@ import { cancelOsDownload } from "Core/update/requests"
 import { URL_ONBOARDING } from "Core/__deprecated__/renderer/constants/urls"
 import { useDeactivateDeviceAndRedirect } from "Core/overview/components/overview-screens/pure-overview/use-deactivate-device-and-redirect.hook"
 import { getDevicesSelector } from "Core/device-manager/selectors/get-devices.selector"
-import { setSelectDeviceDrawerOpen } from "Core/device-select/actions/set-select-device-drawer-open.action"
 
 export const useDeviceDetachedEffect = () => {
   const history = useHistory()
@@ -35,10 +34,6 @@ export const useDeviceDetachedEffect = () => {
     return registerDeviceDetachedListener(
       async (properties: DeviceBaseProperties) => {
         dispatch(removeDevice(properties))
-
-        if (devices.length < 2) {
-          dispatch(setSelectDeviceDrawerOpen(false))
-        }
 
         if (activeDeviceId !== properties.id) {
           return
--- a/libs/core/core/hooks/use-device-detached-effect.ts
+++ b/libs/core/core/hooks/use-device-detached-effect.ts
@@ -16,7 +16,9 @@ import { isActiveDeviceProcessingSelector } from "Core/device-manager/selectors/
 import { deactivateDevice } from "Core/device-manager/actions/deactivate-device.action"
 import { cancelOsDownload } from "Core/update/requests"
 import { URL_ONBOARDING } from "Core/__deprecated__/renderer/constants/urls"
-import { useHandleActiveDeviceDetached } from "Core/overview/components/overview-screens/pure-overview/use-handle-active-device-detached.hook"
+import { useRedirectOnActiveDeviceDetached } from "Core/overview/components/overview-screens/pure-overview/use-redirect-on-active-device-detached.hook"
+import { getDevicesSelector } from "Core/device-manager/selectors/get-devices.selector"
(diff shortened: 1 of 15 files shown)


libs/core/device-initialization/actions/initialize-mudita-pure.action.ts
```diff
@@ -36,72 +36,94 @@ export const initializeMuditaPure = createAsyncThunk<
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const dispatchWithDeviceCheck = async (action: any): Promise<any> => {
+      if (!isActiveDeviceAttachedSelector(getState())) {
+        dispatch(
+          setDeviceInitializationStatus(DeviceInitializationStatus.Aborted)
+        )
+        throw new AppError(DeviceInitializationError.ActiveDeviceNotAttached)
+      }
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+      return dispatch(action)
```
patryk-sierzega, line 48: please remove 'any' and eslint-disables

patryk-sierzega: and second, do we really need to check if device is attached every time we do some reading action from device? Sounds like overkill, but maybe there is some scenario for it.

libs/core/core/hooks/use-device-detached-effect.ts
```diff
@@ -26,13 +28,18 @@ export const useDeviceDetachedEffect = () => {
...
+  const redirectOnActiveDeviceDetached = useRedirectOnActiveDeviceDetached()
 
   useEffect(() => {
     return registerDeviceDetachedListener(
       async (properties: DeviceBaseProperties) => {
         dispatch(removeDevice(properties))
 
+        if (devices.length < 2) {
+          dispatch(setSelectDeviceDrawerOpen(false))
+        }
```
patryk-sierzega, line 41: I've noticed that when executing method 'redirectOnActiveDeviceDetached()' there are similar conditions inside it, won't we do the same things?

libs/core/overview/components/overview-screens/pure-overview/use-redirect-on-active-device-detached.hook.ts
```diff
@@ -7,30 +7,28 @@ import { useHistory } from "react-router-dom"
...
+    if (devices.length > 1) {
+      dispatch(setSelectDeviceDrawerOpen(false))
+      history.push(URL_DISCOVERY_DEVICE.availableDeviceListModal)
+    } else if (devices.length === 1) {
+      history.push(URL_DISCOVERY_DEVICE.root)
+    } else {
+      history.push(URL_MAIN.news)
     }
-  }, [devices, history, dispatch, activeDeviceAttached])
+  }, [devices, history, dispatch])
```
patryk-sierzega, line 33: This hook not only redirects as name says, but also deactivated device. Maybe it would be good idea to rename it to 'useDeactivateDeviceAndRedirect' or similar?

dkarski: Yes, you're correct... I've been thinking about it, and since there are two of us, I'm going to make a change.

libs/core/device-initialization/actions/initialize-mudita-pure.action.ts
dkarski: > please remove 'any' and eslint-disables

I understand the request to remove 'any' and eslint-disables. However, I'm finding it challenging to manage types for this helper in a straightforward manner. Therefore, I've opted for a compromise to avoid overengineering with types, especially since this helper is confined within the action context and doesn't extend beyond it."

dkarski: > and second, do we really need to check if device is attached every time we do some reading action from device? Sounds like overkill, but maybe there is some scenario for it.

I understand your perspective and have been pondering how to handle this differently myself. Perhaps I'm missing something or not seeing the full picture. Could we have a brief sync on this topic? If I explain the issue in detail, it might spark a new idea for me, or you might be able to offer some valuable advice.

libs/core/overview/components/overview-screens/pure-overview/use-redirect-on-active-device-detached.hook.ts
dkarski: https://github.com/mudita/mudita-center/pull/1687/commits/d03a67f37a9df873f450a04e4389a6757bc89eaf

libs/core/core/hooks/use-device-detached-effect.ts
dkarski: after sync https://github.com/mudita/mudita-center/pull/1687/commits/9183a4c36a55af25906ec1da794195de795ae77e

libs/core/device-initialization/actions/initialize-mudita-pure.action.ts
dkarski: After sync review, refactored code to eliminate abstractions and use if statements before invoking async thunks
https://github.com/mudita/mudita-center/pull/1687/commits/f79796a8594fad0ab72bd15b2a3226b29abcaeaa

libs/core/device-initialization/actions/initialize-mudita-pure.action.ts
```diff
@@ -39,20 +39,44 @@ export const initializeMuditaPure = createAsyncThunk<
...
       setDeviceInitializationStatus(DeviceInitializationStatus.Initializing)
     )
+    if (!isActiveDeviceAttachedSelector(getState())) {
+      dispatch(
+        setDeviceInitializationStatus(DeviceInitializationStatus.Aborted)
+      )
+      return DeviceInitializationStatus.Aborted
+    }
+
     await dispatch(loadSettings())
```
patryk-sierzega, line 49: got an idea for it :D 

- in every async action necessary to be checked add this check if !isActiveDeviceAttachedSelector(getState()) and if this true then return rejectWithValue
- in this action 'initializeMuditaPure' check result if inner action returned reject then also reject it
- in place where initializeMuditaPure is invoked check if you got rejected and if yes then execute code below:
`dispatch(
        setDeviceInitializationStatus(DeviceInitializationStatus.Aborted)
      )
      return DeviceInitializationStatus.Aborted`
- if there are few places of invoking initializeMuditaPure then wrap it into function/thunk
      
 In my honest opinion, checking it everytime is overkill, I would suggest to simply reject in every async thunk when something is wrong and then act only when there is a problem. You are not counteracting problems, you simply break execution of this main thunk. So it will be the same.

patryk-sierzega (approved):

libs/core/discovery-device/components/available-device-list.container.tsx
```diff
@@ -19,6 +19,10 @@ const AvailableDeviceListContainer: FunctionComponent = () => {
   const dispatch = useDispatch<Dispatch>()
   const devices = useSelector(getDevicesSelector)
 
+  useEffect(() => {
+    dispatch(setDiscoveryStatus(DiscoveryStatus.Discovering))
```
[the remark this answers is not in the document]
dkarski: my bad, thanks <3 

https://github.com/mudita/mudita-center/pull/1687/commits/432306580ad1c0f47cce962379e07c1714d0dea3

libs/core/overview/components/overview-screens/pure-overview/use-handle-active-device-aborted.hook.ts
```diff
@@ -23,11 +23,13 @@ export const useHandleActiveDeviceAborted = () => {
   const dispatch = useDispatch<Dispatch>()
   const devices = useSelector(getDevicesSelector)
 
-  return useCallback(() => {
-    void dispatch(deactivateDevice())
+  return useCallback(async () => {
+    const pathname = history.location.pathname
+
+    await dispatch(deactivateDevice())
```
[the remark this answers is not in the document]
dkarski: you are right https://github.com/mudita/mudita-center/pull/1687/commits/a5cfa4226c92031ab3ae6a550a25eda0ad1ad8a1

PR #1760, opened 2024-03-06
[CP-2487] Last backup date

lkowalczyk87:
JIRA Reference: [CP-2487]

### :memo: Description ️

-

[CP-2487]: https://appnroll.atlassian.net/browse/CP-2487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/generic-view/store/src/lib/backup/refresh-backup-list.action.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    if (!backupsList.ok) {
+      return rejectWithValue(undefined)
+    }
+
+    const backups =
+      (backupsList.data
+        .map((item) => {
+          const isFormatValid =
+            /^\d+[_][a-zA-Z0-9]+[.]mcbackup$/.exec(item) !== null
```
mkurczewski, line 38: Nothing important but `test` seems to fit little bit more
```suggestion
          const isFormatValid = /^\d+[_][a-zA-Z0-9]+[.]mcbackup$/i.test(item)
```

libs/generic-view/store/src/lib/backup/refresh-backup-list.action.ts
```diff
@@ -0,0 +1,61 @@
...
+
+          if (!isFormatValid) {
+            return null
+          }
+
+          const [fileName] = item.split(".")
+          const [timestamp, serialNumber] = fileName.split("_")
+          const result: Backup = {
+            date: new Date(+timestamp),
```
mkurczewski, line 47: Is this `+` some kind of converting it to number? If so, i think `Number` would be much cleaner to read
```suggestion
            date: new Date(Number(timestamp)),
```

lkowalczyk87: `+` is equivalent to `Number`

mkurczewski (approved):

PR #1763, opened 2024-03-07
[CP-2555] mudita-center-e2e Framework ignores chromedriver arguments  --headless

dkarski:
JIRA Reference: [CP-2555]

### :memo: Description ️

- Implemented headless E2E test support on Linux using Xvfb, ensuring compatibility with CI/CD on GitHub Actions.
- Updated documentation with instructions for running tests in headless mode.
- Introduced a filter based on the CI environment variable, allowing for selective execution of tests suitable for the CI/CD environment.
- Added a Jira task for extending headless mode support to Mac and Windows, expanding the scope of E2E testing.

[CP-2555]: https://appnroll.atlassian.net/browse/CP-2555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/mudita-center-e2e/wdio.conf.ts
```diff
@@ -55,29 +55,7 @@ export const config: Options.Testrunner = {
...
-    toRelativePath(TestFilesPaths.messagesInAppNavigationTest),
-    toRelativePath(TestFilesPaths.newsInAppNavigationTest),
-    toRelativePath(TestFilesPaths.overviewInAppNavigationTest),
-    toRelativePath(TestFilesPaths.settingsInAppNavigationTest),
-    toRelativePath(TestFilesPaths.helpWindowCheckTest),
-    toRelativePath(TestFilesPaths.sarWindowCheckTest),
-    toRelativePath(TestFilesPaths.tosPrivacyLicenceWindowsCheckTest),
-    toRelativePath(TestFilesPaths.messageSendTest),
-  ],
-  suites: {
```
tomaszmaleckimudita, line 74: You deleted main functionality of automation which is suites for testing various things:

- smoke test set
- mc standalone tests
- harmony only tests
- pure only tests
- kompakt only tests 
- multidevice tests

apps/mudita-center-e2e/wdio.conf.ts
```diff
@@ -115,10 +91,9 @@ export const config: Options.Testrunner = {
       // maxInstances: 5,
       //
       browserName: "chrome",
-      // acceptInsecureCerts: true,
       "goog:chromeOptions": {
         binary: process.env.TEST_BINARY_PATH,
-        args: ["--headless", "--no-sandbox"],
+        args: [],
```
tomaszmaleckimudita, line 96: what about other arguments? Task is indicating that none of the arguments are used. --headless is understandable, but what about --no-sandbox and others?

tomaszmaleckimudita (changes requested):
/src/suites.constant.ts is adding unnessesary work. Can we modify or move test-filenames.const.ts so we need to edit only one file when adding new test? 

Suites:{} object needs to be supported in some way - so we can add test sets:
- smoke test set
- mc standalone tests
- harmony only tests
- pure only tests
- kompakt only tests
- multidevice tests

apps/mudita-center-e2e/wdio.conf.ts
dkarski: I need to divide the test set into those that run locally vs. those that are to be run on CI/CD.

The file can support suites, if you want, it's not a problem. Do you already have some suites defined so that I can take them into account?

apps/mudita-center-e2e/wdio.conf.ts
dkarski: You don't need the --no-sandbox and other arguments to run the tests locally or on CI/CD, application. Do you have another use case where you need to pass args? Those you're pointing out are not needed to properly run the tests.

apps/mudita-center-e2e/wdio.conf.ts
dkarski: https://github.com/mudita/mudita-center/pull/1763/commits/f510db0579f2e85c73ada897ab6bfaf043cb0b74

tomaszmaleckimudita (approved):

PR #1819, opened 2024-04-08
[CP-2701] Added search input component

mkurczewski:
JIRA Reference: [CP-2701]

### :memo: Description ️

-

[CP-2701]: https://appnroll.atlassian.net/browse/CP-2701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/generic-view/ui/src/lib/interactive/input/search-input.tsx
```diff
@@ -0,0 +1,136 @@
...
+  label: string
+}
+
+export const SearchInput: APIFC<Data, Config> = ({ data, config }) => {
+  const id = useId()
+  const { register, watch, setValue } = useFormContext()
+  const value = (watch(config!.name) as string) || ""
+
+  const clear = () => {
+    setValue(config!.name, "")
```
dkarski, line 30: ⚪ (question)

Is using `!` with `config.name` necessary, and is there a safer alternative? Concerned about potential crashes

dkarski (approved):

lkowalczyk87 (approved):

libs/generic-view/ui/src/lib/interactive/input/search-input.tsx
mkurczewski: This is because of `config` being optional (only as a type), while in fact it isn't. External API for generic views will prevent this but not placing a "name" or even a "config" should actually crash the app as it's just an error.

However, I'll add a topic for our front-end meeting so maybe we can discuss better solution from TypeScript's perspective.

PR #1823, opened 2024-04-09
[CP-2705] map google contacts to unified contact

lkowalczyk87:
JIRA Reference: [CP-2705]

### :memo: Description ️

-

### DIFF
--- a/libs/device/models/src/lib/import-contacts/unified-contact.ts
+++ b/libs/device/models/src/lib/import-contacts/unified-contact.ts
@@ -7,11 +7,37 @@ export type UnifiedContact = {
   id: string
   firstName?: string
   lastName?: string
+  middleName?: string
+  honorificPrefix?: string
+  honorificSuffix?: string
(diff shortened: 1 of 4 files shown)


libs/device/models/src/lib/import-contacts/unified-contact.ts
```diff
@@ -7,11 +7,37 @@ export type UnifiedContact = {
...
   }[]
+  emailAddresses: {
+    value: string
+    type?: string
+    preference?: number
+  }[]
+  addresses: {
+    type?: string
+    street?: string
+    streetExtended?: string
```
dkarski, line 28: 🟢 

Hey, sorry I didn't catch this during our sync. Could we change the field name from streetExtended? to extendedStreet?? I believe this change will bring more consistency to our field naming conventions. What do you think?

```suggestion
    extendedStreet?: string
```

dkarski (approved):

PR #1832, opened 2024-04-11
[CP-2720] Modal refactor (dot notation)

mkurczewski:
JIRA Reference: [CP-2720]

### :memo: Description ️

-


libs/generic-view/ui/src/lib/interactive/modal/helpers/modal-size-controller.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const ModalSizeControllerMedium = styled.div`
+  display: none;
+`
+
+export const ModalSizeControllerLarge = styled.div`
+  display: none;
+`
+
+export const ModalSizeController: APIFC<undefined, Config> = ({ config }) => {
```
dkarski, line 28: ⚪ (question)
s this component a placeholder for when designs for other sizes are introduced? From my understanding, it currently doesn’t perform any function, correct?

dkarski (approved):
I've encountered dot notation before, often misused or overdone. Here, I believe this convention fits perfectly for a component that defines a domain-specific group of sub-components you can utilize.

This approach reduces the number of imports and enhances the readability of the component and its custom derivatives.

Great job on this implementation!

❤️ ❤️ ❤️

libs/generic-view/ui/src/lib/interactive/modal/helpers/modal-size-controller.tsx
dkarski: I see: https://github.com/mudita/mudita-center/pull/1832/files#diff-eb75d6ebce3f39c3f0a2bb8ddd2e35b9956330d5c4aefc05f241f19635174657R53

dkarski: Why is the modal size specified by a dedicated component instead of props? If it's more convenient for you, let's catch up on a sync. Maybe you could help me understand this idea.

PR #1862, opened 2024-04-25
[CP-2732] Import contacts from VCF

mkurczewski:
JIRA Reference: [CP-2732]

### :memo: Description ️

-

[CP-2732]: https://appnroll.atlassian.net/browse/CP-2732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/generic-view/store/src/lib/imports/contacts-mappers/vcard/map-vcard.ts
```diff
@@ -0,0 +1,152 @@
...
+    })
+    .filter(({ value }) => value)
+}
+
+const getId = (item: jCard) => {
+  const field = first(getFields(item, "uid"))
+  return field?.value as string | undefined
+}
+
+const getN = (
```
lkowalczyk87, line 54: I think we need a more explicit name for this function

lkowalczyk87 (approved):

libs/generic-view/store/src/lib/imports/contacts-mappers/vcard/map-vcard.ts
mkurczewski: Changed

PR #1946, opened 2024-06-28
[CP-2835] Added support for cancelling data migration process

mkurczewski:
JIRA Reference: [CP-2835]

### :memo: Description ️

-

[CP-2835]: https://appnroll.atlassian.net/browse/CP-2835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lkowalczyk87 (approved):

libs/generic-view/feature/src/lib/modals/data-migration-error-modal.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const dataMigrationStatus = useSelector(selectDataMigrationStatus)
+  const opened = dataMigrationStatus === "FAILED" && !activeDevice
+
+  const onClose = () => {
+    dispatch(setDataMigrationStatus("IDLE"))
+    dispatch(setDataMigrationProgress(DataMigrationPercentageProgress.None))
+    dispatch(setDataMigrationFeatures([]))
+    dispatch(setDataMigrationSourceDevice(undefined))
+    dispatch(setDataMigrationPureDbIndexing(false))
+    dispatch(setDeviceErrorModalOpened(false))
```
dkarski, line 34: ⚪ (question)
Is this the only place where the state is reset? Maybe we should consolidate this group of actions into one place?

dkarski (approved):

libs/generic-view/feature/src/lib/modals/data-migration-error-modal.tsx
mkurczewski: There are more places like that but depending on context, different actions are dispatched so I actually switched from single thunk to multiple dispatches because of it ;)

PR #1950, opened 2024-07-02
[CP-2936] Refactor Controllers to Services name convention

dkarski:
JIRA Reference: [CP-2936]

### :memo: Description ️

-

[CP-2936]: https://appnroll.atlassian.net/browse/CP-2936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mkurczewski (approved):

libs/device-protocol/feature/src/services/serial-port.service.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import SerialPort from "serialport"
+
+export interface ISerialPortService {
+  list(): Promise<SerialPort.PortInfo[]>
+}
+
+export class SerialPortService implements ISerialPortService {
+  list(): Promise<SerialPort.PortInfo[]> {
+    return SerialPort.list()
```
lkowalczyk87, line 14: it probably should use mock devices list when mock service is enabled

dkarski: https://github.com/mudita/mudita-center/pull/1950/files#diff-73c889eb7b000757a16fa2a10d50a2826b9b0253fe584bc0b47f1e0f4b0b32c9R157

PR #1977, opened 2024-07-22
CP-2987 Fix order of tests as well as mock vs non-contract tests + Fix linux tests 

tomaszmaleckimudita:
JIRA Reference: [CP-2987]

### :memo: Description ️
* CI/CD split into mock and non-contract tests
* Mock tests should contain offline tests if then need to (verify that, because it looks like they work on non-contract build https://appnroll.atlassian.net/browse/CP-2880
* Standalone should not contain any offline tests (mock) 
* Make sure there is no missing tests in specs 
* Add npm runs to package.json and /mudita-center-e2e/package.json
* Test have issues with extracted appimage on linux

[CP-2987]: https://appnroll.atlassian.net/browse/CP-2987?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/mudita-center-e2e/wdio.conf.ts
```diff
@@ -113,6 +110,11 @@ export const config: Options.Testrunner = {
       toRelativePath(TestFilesPaths.helpWindowCheckOfflineTest),
       toRelativePath(TestFilesPaths.kompaktOverview),
```
tomaszmaleckimudita, line 111: Remove this one

robertmudi (dismissed):

PR #2034, opened 2024-08-21
[CP-2801] Back and Forward mouse buttons cause unexpected MDS behavior

dkarski:
JIRA Reference: [CP-2801]

### :memo: Description ️

-

[CP-2801]: https://appnroll.atlassian.net/browse/CP-2801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/core/core/components/apps/help-app.component.tsx
```diff
@@ -23,6 +24,7 @@ const getStoreData = async (key?: string) =>
 
 const HelpApp: FunctionComponent = () => {
   useAltLinkDownloadPreventer()
+  useBackForwardButtonPreventer()
```
mkurczewski, line 27: This file is going to be removed in https://github.com/mudita/mudita-center/pull/2033 so you can revert those changes to avoid conflicts.

libs/core/core/components/use-back-forward-button-preventer.hook.tsx
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright (c) Mudita sp. z o.o. All rights reserved.
+ * For licensing, see https://github.com/mudita/mudita-center/blob/master/LICENSE.md
+ */
+
+import { useCallback, useEffect } from "react"
+
+const useBackForwardButtonPreventer = () => {
+  const handleMouseUp = useCallback((event: MouseEvent) => {
+    if (event.button === 3 || event.button === 4) {
```
mkurczewski, line 10: Are you sure those values will always catch the prev & next buttons? What about gaming mouses etc.?

dkarski: Values 3 and 4 are standard and typically handle the 'back' and 'forward' buttons in most cases, according to the MDN documentation. Even with gaming mouses, which might allow remapping, standard browsers generally still recognize these buttons according to this standard. More details can be found in the MDN documentation: [MouseEvent.button](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button). 

I don't have a gaming mouse, so I'm basing my response on the documentation and information available online. If you have such a mouse, I can conduct a test to verify this.

MateuszMudita (approved):

libs/core/core/components/use-back-forward-button-preventer.hook.tsx
mkurczewski: Ok, that should be enough 👍

mkurczewski (approved):

PR #2076, opened 2024-09-23
[CP-3004] Refactored button action to accept multiple actions

mkurczewski:
JIRA Reference: [CP-3004]

### :memo: Description ️

Refactored button's `action` property to accept an array of actions that can be executed one by one at single click.

[CP-3004]: https://appnroll.atlassian.net/browse/CP-3004?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/generic-view/feature/src/lib/setup-component.tsx
```diff
@@ -79,17 +79,15 @@ export const setupComponent = <P extends object>(
...
 
-    const editableProps = {
-      ...dataProps,
-      config: {
-        ...config,
-      },
-      data: undefined as unknown,
-    }
+    const editableProps = JSON.parse(
+      JSON.stringify({
```
dkarski, line 83: Hey! I noticed you're using `JSON.stringify/parse` here — are you aiming for a deep copy? If so, maybe `_.cloneDeep()` would be clearer and more predictable in terms of intent. What do you think?

libs/generic-view/feature/src/lib/setup-component.tsx
```diff
@@ -163,6 +169,38 @@ export const setupComponent = <P extends object>(
   }
 }
 
+const processFormFields = (
+  field: DataProviderExtendedField,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  value: any
```
dkarski, line 175: Maybe replace `any` with `unknown` and use union type checks (e.g., `is List`, `is Boolean`) for safer type checking?

dkarski (commented):
The change looks good to me. I just see two small things to address.

Please also ensure that it doesn’t affect the configuration coming from Kompakt or the existing e2e tests. If it does, I suggest making the necessary changes in Kompakt and updating the e2e tests as part of this PR before merging.

libs/generic-view/feature/src/lib/setup-component.tsx
mkurczewski: Good point. Changed in [34c1131](https://github.com/mudita/mudita-center/pull/2076/commits/34c113113a5c96cbee59a520d9a1f43401af6e57)

libs/generic-view/feature/src/lib/setup-component.tsx
mkurczewski: Fixed in [`c8028ba` (#2076)](https://github.com/mudita/mudita-center/pull/2076/commits/c8028ba6b8efa5786ed2b360909d7f6c983e4cea)

dkarski (approved):

PR #2087, opened 2024-09-26
[CP-3004] Added helper components for contacts list selection

mkurczewski:
JIRA Reference: [CP-3004]

### :memo: Description ️

-

[CP-3004]: https://appnroll.atlassian.net/browse/CP-3004?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/generic-view/ui/src/lib/interactive/format-message.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { selectActiveDeviceConfiguration } from "generic-view/store"
+
+export const FormatMessage: APIFC<FormatMessageData, FormatMessageConfig> = ({
+  data,
+  config,
+  children,
+  ...props
+}) => {
+  const id = useId()
+  const deviceConfig = useSelector(selectActiveDeviceConfiguration)
```
dkarski, line 25: ⚪ 
Is it necessary to tie this component to the active device state? Should lang come from the API or a default configuration Provider instead?

libs/generic-view/ui/src/lib/interactive/modal/helpers/modal-title-icon.tsx
```diff
@@ -14,6 +14,7 @@ export const ModalTitleIcon: APIFC<undefined, ModalTitleIconConfig> = ({
   children,
   ...rest
 }) => {
+  // @ts-ignore
```
dkarski, line 17: Why was this @ts-ignore added? I removed it locally, and TypeScript doesn't throw any errors 🤷‍♂️

dkarski (approved):

libs/generic-view/ui/src/lib/interactive/modal/helpers/modal-title-icon.tsx
mkurczewski: It doesn't throw until... it throws. Sometimes during hot-reload you can get TS error from here in the console, while in IDE everything works fine 🤷‍♂️ . I tried to fix it but couldn't find the real issue, so I added the ignore command to not be forced to restart the app every time it happens.

libs/generic-view/ui/src/lib/interactive/format-message.tsx
mkurczewski: For now only components are generated only in context of active device.

PR #2126, opened 2024-10-14
[CP-3185] Integrate the flashing mechanism with the application on Windows

dkarski:
JIRA Reference: [CP-3185]

[CP-3185]: https://appnroll.atlassian.net/browse/CP-3185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/msc-flash/msc-flash-harmony/src/lib/selectors/select-is-flashing-in-state.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { createSelector } from "@reduxjs/toolkit"
+import { FlashingProcessState } from "../constants"
+import { selectFlashingProcessState } from "./select-flashing-process-state"
+
+export const selectIsFlashingInState = (targetState: FlashingProcessState) =>
+  createSelector(
+    selectFlashingProcessState,
+    (flashingProcessState) => flashingProcessState === targetState
+  )
```
mkurczewski, line 14: I'm not sure this is needed. You can do the same by using the already existing selector `selectFlashingProcessState`. Creating a specialized selector for such comparison doesn't make sense unless you're going to use it hundreds of times and with the same `targetState` argument.

Notice how the second approach is much easier and more flexible to use:
```TSX
const isDownloadingFiles = useSelector((state: ReduxRootState) => { 
  return selectIsFlashingInState(FlashingProcessState.DownloadingFiles)
})
const isUnpackingFiles = useSelector((state: ReduxRootState) => { 
  return selectIsFlashingInState(FlashingProcessState.UnpackingFiles)
})

// ...

if (isDownloadingFiles) {
 // do something
} else if (isUnpackingFiles) {
  // do something
}
```
vs
```TSX
const flashingState = useSelector(selectFlashingProcessState)
const isDownloadingFiles = flashingState === FlashingProcessState.DownloadingFiles
const isUnpackingFiles = flashingState === FlashingProcessState.UnpackingFiles

if (isDownloadingFiles) {
 // do something
} else if (isUnpackingFiles) {
  // do something
}

// it also allows to use switch instead

switch (flashingState) {
  case FlashingProcessState.DownloadingFiles:
	// do something
	break
  case FlashingProcessState.UnpackingFiles:
	// do something
	break
}
```

dkarski: I created this selector mainly for memoization – when the Redux state doesn't change, the result is cached, avoiding unnecessary recalculations. Your approach is simpler, but my goal was to prevent extra computations.

mkurczewski (approved):

PR #2128, opened 2024-10-15
[CP-3170] Minor refactor of device-flash

dkarski:
JIRA Reference: [CP-3170]

### :memo: Description ️

-

[CP-3170]: https://appnroll.atlassian.net/browse/CP-3170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/core/core/hooks/use-msc-device-detached-effect.ts
```diff
@@ -0,0 +1,55 @@
...
+  DeviceType,
+} from "device-protocol/models"
+import {
+  abortMscFlashing,
+  FlashingProcessState,
+  selectIsFlashingInActivePhases,
+} from "msc-flash-harmony"
+import { Dispatch } from "Core/__deprecated__/renderer/store"
+
+export const useMscDeviceDetachedEffect = () => {
```
mkurczewski, line 21: Why another hook if we already have the `useDeviceDetachedEffect` ?

dkarski: Initially, I wanted to handle this code in a single hook, but due to circular dependencies, I had to add a new hook related to detach. I suggest leaving it as is and creating a technical task where we can merge the three hooks into one and resolve the dependency issue.

mkurczewski: In case of dependencies issues you can try an import like "../../../../../../../........"

dkarski: https://appnroll.atlassian.net/browse/CP-3194

mkurczewski (approved):

PR #2154, opened 2024-10-29
[CP-3205] Recovery Completed modal

MateuszMudita:
JIRA Reference: [CP-3205]

### :memo: Description ️

-

[CP-3205]: https://appnroll.atlassian.net/browse/CP-3205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/core/__deprecated__/renderer/components/core/icon/icon.config.ts
```diff
@@ -508,6 +509,7 @@ const typeToIcon: Partial<Record<IconType, typeof Arrow>> = {
   [IconType.RecoveryModeWhite]: RecoveryModeWhite,
   [IconType.RecoveryModeBlack]: RecoveryModeBlack,
   [IconType.ButtonSuccess]: ButtonSuccess,
+  [IconType.LightButton]: LightButton,
```
dkarski, line 512: ⚪ (project sugesstion)

What do you think about using a transparent background for better versatility?

A transparent background might help the icon adapt better across different backgrounds, keeping a cleaner visual consistency.

libs/core/core/hooks/use-msc-device-detached-effect.ts
```diff
@@ -40,6 +41,9 @@ const useHandleDevicesDetached = () => {
   const flashingInWaitingForBackButtonState = useSelector(
     selectIsFlashingInState(FlashingProcessState.WaitingForBackButton)
   )
+  const flashingInRestartingPhase = useSelector(
+    selectIsFlashingInState(FlashingProcessState.Restarting)
+  )
```
dkarski, line 46: Hey, I wanted to point out this code because I made a mistake in my previous comment. Maybe we could fix it together by using a selector with dynamic arguments instead of a factory? This way, we can avoid creating a new selector with each re-render, allowing it to retain the memorized values.

```suggestion
  const flashingInWaitingForBackButtonState = useSelector((state) => 
    selectIsFlashingInState(state, FlashingProcessState.WaitingForBackButton)
  )
  const flashingInRestartingPhase = useSelector((state) => 
    selectIsFlashingInState(state, FlashingProcessState.Restarting)
  )
```

dkarski (approved):
Approved for now – let me know if you need help with linting!

PR #2273, opened 2025-01-15
[CP-3378] Fix error with wrong number of deleted files

MateuszMudita:
JIRA Reference: [CP-3378]

### :memo: Description ️

-

[CP-3378]: https://appnroll.atlassian.net/browse/CP-3378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/generic-view/ui/src/lib/table/table.tsx
```diff
@@ -104,11 +100,14 @@ export const Table: APIFC<TableData, TableConfig> & {
...
-        formContext.setValue(
-          formOptions.selectedIdsFieldName,
-          intersection(data, unavailableIds)
-        )
+        setTimeout(() => {
+          formContext.setValue(
+            formOptions.selectedIdsFieldName!,
+            intersection(data, unavailableIds)
+          )
+        }, 500)
```
dkarski, line 110: Is it necessary? What is the source of the problem? Can this issue be resolved at the source?

libs/generic-view/ui/src/lib/table/table.tsx
```diff
@@ -104,11 +100,14 @@ export const Table: APIFC<TableData, TableConfig> & {
         formOptions.selectedIdsFieldName
       )
       const unavailableIds = difference(selectedIds, data)
+
       if (unavailableIds.length > 0) {
```
dkarski, line 104:
```suggestion
      if (unavailableIds.length > 0 && selectedIdsFieldName !== undefined)
```

libs/generic-view/ui/src/lib/table/table.tsx
```diff
@@ -104,11 +100,14 @@ export const Table: APIFC<TableData, TableConfig> & {
...
-        formContext.setValue(
-          formOptions.selectedIdsFieldName,
-          intersection(data, unavailableIds)
-        )
+        setTimeout(() => {
+          formContext.setValue(
+            formOptions.selectedIdsFieldName!,
```
dkarski, line 107:
```suggestion
            formOptions.selectedIdsFieldName,
```

libs/generic-view/ui/src/lib/table/table.tsx
MateuszMudita: Unfortunately this is source of problem. Sometimes toast is rendered after `selectedIdsFieldName` is cleared, so we have to add setTimeout

libs/generic-view/ui/src/lib/table/table.tsx
MateuszMudita: It won't work this way (TypeScript still thinks that this field can be `undefined`. I've added if statement inside of setTimeout instead.

dkarski (approved):

libs/generic-view/ui/src/lib/table/table.tsx
mkurczewski: Maybe it'd be better to export `toastAnimationDuration` const from `toast.tsx` file and use this value instead?

PR #2275, opened 2025-01-15
[CP-3369] Error handling during file delete

MateuszMudita:
JIRA Reference: [CP-3369]

### :memo: Description ️

-

[CP-3369]: https://appnroll.atlassian.net/browse/CP-3369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/core/__deprecated__/renderer/locales/default/en-US.json
```diff
@@ -1014,6 +1014,10 @@
   "module.genericViews.dataMigration.failure.partialChangesDescription": "Some data transferred, but you'll need to check Kompakt to see what's missing.",
   "module.genericViews.dataMigration.failure.notEnoughSpace": "Your phone’s storage is full.",
   "module.genericViews.dataMigration.failure.closeButtonLabel": "Close",
+  "module.genericViews.entities.delete.failure.all.modalTitle": "Deleting failed",
+  "module.genericViews.entities.delete.failure.some.modalTitle": "Deleting some files failed",
```
dkarski, line 1018: Shouldn't singular and plural handling be necessary in the translation here? 🤔

libs/generic-view/models/src/lib/entities-delete-error.ts
```diff
@@ -0,0 +1,23 @@
...
+import { z } from "zod"
+
+const dataValidator = z.undefined()
+
+const configValidator = z.object({
+  modalKey: z.string(),
+  entitiesType: z.string(),
+})
+
+// export type EntitiesDeleteError = z.infer<typeof dataValidator>
```
dkarski, line 15: keep or not keep 😅 
```suggestion
export type EntitiesDeleteError = z.infer<typeof dataValidator>
```
or
```suggestion
```

dkarski (dismissed):
✅  PR is fine, just two small comments to check if we’re leaving it like this.

libs/core/__deprecated__/renderer/locales/default/en-US.json
MateuszMudita: No. By design we have only these two modal titles :)

libs/generic-view/models/src/lib/entities-delete-error.ts
MateuszMudita: Forget to remove it 😅

PR #2291, opened 2025-01-27
API configuration testing - test case to check if the response data i…

slawomir-werner:
JIRA Reference: [CP-3421]

### :memo: Description ️

- API configuration testing - test case to check if the response data is valid, 
- Fixed a bug where Mudita Center couldn't connect to the Kompakt in API testing

[CP-3421]: https://appnroll.atlassian.net/browse/CP-3421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/api-devices-testing/src/lib/api-configuration.spec.tsx
+++ b/libs/api-devices-testing/src/lib/api-configuration.spec.tsx
@@ -6,6 +6,11 @@
 import { DeviceProtocol } from "device-protocol/feature"
 import { setKompaktConnection } from "./helpers/set-connection"
 import { APIConfigService } from "device/feature"
+import {
+  ApiConfig,
+  GeneralError
+} from "device/models"
+import exp from "constants"
 
 jest.mock("shared/utils", () => {
   return { callRenderer: () => {} }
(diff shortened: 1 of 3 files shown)


libs/api-devices-testing/src/lib/api-configuration.spec.tsx
```diff
@@ -47,4 +51,48 @@ describe("API configuration", () => {
...
+    expect(apiConfig.osVersion).toMatch(/^MuditaOS K/)
+    expect(apiConfig.lang).toMatch(/^[a-z]{2}-[A-Z]{2}$/)
+    expect(apiConfig.variant?.length).toBeGreaterThan(0)
+    expect(apiConfig.features.sort()).toEqual(["mc-overview","contacts","mc-data-migration","fileManager"].sort())
+    expect(apiConfig.entityTypes?.sort()).toEqual(["contacts","audioFiles","imageFiles","ebookFiles"].sort())
+    expect(apiConfig.productId).toEqual("2006")
+    expect(apiConfig.vendorId).toEqual("0e8d")
+    expect(apiConfig.serialNumber).toMatch(/^[A-Z0-9]{13}$/)
+    expect(apiConfig.otaApiConfig?.otaApiKey.length).toEqual(15)
+    expect(apiConfig.otaApiConfig?.osVersionTimestamp.toString().length).toEqual(10)
```
MateuszMudita, line 95: Do we want to have this hardcoded values for this expects? Maybe we should keep it in some constant? What will happen if apiConfig will return new features for example?

MateuszMudita (approved):

PR #2346, opened 2025-02-27
[CP-3175] [API Device][Kompakt] Contacts - searching

robertmudi:
JIRA Reference: [CP-3175]

### :memo: Description ️

-Mock and Connect Kompakt with 10 contacts

Test Contacts search mechanism for various data fields

Test number of search results 

If there are no search results, the user is informed about the fact.

Test Search rules and sorting order

if the user selects a searched contact he moves to the contact details, the search field is cleared and the selected contact is highlighted in the contact list.

[CP-3175]: https://appnroll.atlassian.net/browse/CP-3175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/mudita-center-e2e/src/specs/overview/kompakt-contacts-search.ts
```diff
@@ -0,0 +1,211 @@
...
+    // 3.1 check if 17 results will appear
+    const updatedresults = await $$(
+      '//li[@data-testid and starts-with(@data-testid, "ui-form-search-results-")]'
+    )
+    await expect(updatedresults.length).toBeGreaterThan(16) //there are 17 results for this search
+    //3.2 clear search field
+    await searchSuggestionsListEraseButton.click()
+
+    //4.0 search using only phone number
+    const telNum: string = "+48345678902"
```
mkurczewski, line 132:
```suggestion
    const telNum = "+48345678902"
```
This will be fine too as TypeScript will figure out the type from the value itself.

Same for website and email consts.

apps/mudita-center-e2e/src/specs/overview/kompakt-contacts-search.ts
```diff
@@ -0,0 +1,211 @@
...
+      await ContactsKompaktPage.contactDisplayNameHeader
+    await expect(contactDisplayNameHeader).toHaveText("Dr. Anna Nowak Jr.")
+
+    //check if selected contact is highlighted on the contacts list
+
+    const selectedContact = ContactsKompaktPage.allContactsTableRows[7]
+    const backgroundColor = await selectedContact.getCSSProperty(
+      "background-color"
+    )
+    await expect(backgroundColor.value).not.toBe("rgb(251, 251, 251)") // Default background color
```
mkurczewski, line 209: I'm wondering if wouldn't be better to add some "active" states with `data-*` tags. The test will break in case of changes in theme or adding for example a new (dark) theme.
This is not a change request for now, but something to consider.

mkurczewski (approved):

MateuszMudita (approved):

apps/mudita-center-e2e/src/specs/overview/kompakt-contacts-search.ts
robertmudi: unfortunately we do not support the negative scenarios so it should stay as is, but thanks for input

PR #2361, opened 2025-03-12
[CP-3518] testing outbox responses

slawomir-werner:
JIRA Reference: [CP-3518]

### :memo: Description ️

-

[CP-3518]: https://appnroll.atlassian.net/browse/CP-3518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/api-devices-testing/src/lib/outbox.spec.tsx
```diff
@@ -0,0 +1,69 @@
...
+    deviceProtocol = setActiveDevice(await setKompaktConnection())
+  })
+
+  afterEach(async () => {
+    await deviceProtocol.activeDevice?.disconnect()
+  }, 10000)
+
+  it("should return valid outbox response for valid deviceId", async () => {
+    expect(deviceId).toBeDefined()
+    if (deviceId == undefined) {
```
MateuszMudita, line 47: I am not sure if you should check this condition. In previous line (`expect(deviceId).toBeDefined()`) you already have checked if deviceId is defined. If deviceId will be undefined, then your test will fail on expect from line 46. Thats why you don't need this if statement here. 

You can have 2 approaches to this problem: 
1 . You want your test to fail if `deviceId` is undefined. Then you should leave expect from line 46 and everything should be fine. 
2. You don't want your test to fail - remove `expect` and in if statement use 3 equal signs, not 2: 
    `if (deviceId === undefined) {`

libs/api-devices-testing/src/lib/outbox.spec.tsx
```diff
@@ -0,0 +1,69 @@
...
+    expect(result.ok).toBeTruthy()
+    const data = result.data as Outbox
+    expect(data.features).toEqual(expect.any(Array))
+    expect(data.entities).toEqual(expect.any(Array))
+    expect(data.data).toEqual(expect.any(Array))
+  })
+
+  it("should return error outbox response for invalid deviceId", async () => {
+    expect(deviceId).toBeDefined()
+    if (deviceId == undefined) {
```
MateuszMudita, line 61: Same as above

libs/api-devices-testing/src/lib/outbox.spec.tsx
mkurczewski: I assume it's only a typescript hack to make "deviceId" defined for next lines. Maybe it'd be better to just change the line 33 from
```TS
deviceId = deviceProtocol.activeDevice?.id
```

to 

```TS
deviceId = deviceProtocol.activeDevice!.id // or deviceId = deviceProtocol.activeDevice?.id as string
```

mkurczewski (commented):
Nothing more to add

MateuszMudita (approved):

PR #2425, opened 2025-04-28
[CP-3529] I want to be informed that the MTP transfer is not available

dkarski:
JIRA Reference: [CP-3529]

### :memo: Description ️

-

[CP-3529]: https://appnroll.atlassian.net/browse/CP-3529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/app-mtp/src/lib/dotnet-mtp/dotnet-mtp.ts
```diff
@@ -138,7 +157,16 @@ export class DotnetMtp implements MtpInterface {
...
+      this.abortController.signal
     )
+      .then(() => {
+        console.log(`${PREFIX_LOG} command status: finished`)
+      })
+      .catch((error) => {
+        const appError = { type: error } as AppError
+        this.uploadFileTransactionStatus[transactionId].error = appError
+        console.error(`${PREFIX_LOG} command status: error`, error)
+      })
```
mkurczewski, line 170: Why mixing async/await with thenable?

libs/app-mtp/src/lib/dotnet-mtp/utils/handle-command.ts
```diff
@@ -25,7 +26,8 @@ const PREFIX_LOG = `[app-mtp/dotnet-mtp]`
 export async function runCommand(
```
mkurczewski, line 26: Since there's no `this` required inside this function's body, I'd stick to one convention and use the arrow function instead.

libs/app-mtp/src/lib/dotnet-mtp/dotnet-mtp.ts
```diff
@@ -118,10 +120,27 @@ export class DotnetMtp implements MtpInterface {
...
+    if (this.uploadFileTransactionStatus[data.transactionId] !== undefined) {
+      this.abortController?.abort()
+      console.log(
+        `${PREFIX_LOG} Canceling upload for transactionId ${data.transactionId}, signal abort status: ${this.abortController?.signal.aborted}`
+      )
+      return Result.success({})
+    } else {
+      return Result.failed({
+        type: MTPError.MTP_TRANSACTION_NOT_FOUND,
+      } as AppError)
```
mkurczewski, line 135: I wonder if in this case we should return an error-like response when transaction wasn't found. After all, the transaction is gone, and that's exactly what user was expecting.

libs/app-mtp/src/lib/node-mtp/node-mtp.ts
```diff
@@ -108,6 +110,22 @@ export class NodeMtp implements MtpInterface {
...
+    if (this.uploadFileTransactionStatus[data.transactionId] !== undefined) {
+      this.abortController?.abort()
+      console.log(
+        `${PREFIX_LOG} Canceling upload for transactionId ${data.transactionId}, signal abort status: ${this.abortController?.signal.aborted}`
+      )
+      return Result.success({})
+    } else {
+      return Result.failed({
+        type: MTPError.MTP_TRANSACTION_NOT_FOUND,
+      } as AppError)
```
mkurczewski, line 125: Same concerns as for `dotnet-mtp.ts` file.

mkurczewski (approved):

PR #2454, opened 2025-05-19
[CP-3618] Check periodically if the MTP is available

slawomir-werner:
JIRA Reference:[CP-3618]

### :memo: Description ️

-

[CP-3618]: https://appnroll.atlassian.net/browse/CP-3618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dkarski (dismissed):

dkarski (approved):

dkarski (approved):

libs/generic-view/store/src/lib/file-transfer/send-files.action.ts
```diff
@@ -83,114 +84,176 @@ export const sendFiles = createAsyncThunk<
...
+        if (isMtpAvailable) {
+          console.log("MTP became available, switching mode.")
+          sendFileAbortController.abort()
+          switchToMtpCounter++
         }
+      }
+    }, 3_000)
+    let currentFileIndex = 0
+
+    const processFiles = async () => {
```
mkurczewski, line 125: I'd consider moving the `processFiles` to a separate action (async thunk). This would split the logic into smaller files and could help in handling _abort_ state using the native thunk's behavior instead of `wasAborted` variable.

libs/shared/utils/src/lib/delay.ts
```diff
@@ -7,9 +7,36 @@
...
+      cleanup();
+      resolve();
+    };
+
+    const cleanup = () => {
+      signal?.removeEventListener('abort', onAbort);
+    };
+
+    signal?.addEventListener('abort', onAbort);
+  });
```
mkurczewski, line 41: I'm not sure this is valid approach. The `delay` method is widely used, but the new logic is used only in one place. Instead of this, I'd revert the previous implementation and just use it like this:

```TSX
delay(signal.aborted ? 0 : 500)
```

Looking at its implementation...

```TSX
      if (signal.aborted) {
        return new AppError(ApiFileTransferError.Aborted, "Aborted")
      }

      if (!ok) {
        return error
      }

      await delay(500, signal)
```

... you can clearly see, that if signal is aborted, the delay function probably won't be called at all as there's a `return` before, and even if it would be called somehow, the solution proposed above should resolve the issue.

mkurczewski: Discussed and understood 🙂

mkurczewski (approved):

PR #2489, opened 2025-06-06
[CP-XXX] Added `getCurrentDevices` method to app-serialport

mkurczewski:
JIRA Reference: [CP-XXX]

### :memo: Description ️

-

### DIFF
--- a/libs/app-serialport/devices/src/lib/serial-port-device.ts
+++ b/libs/app-serialport/devices/src/lib/serial-port-device.ts
@@ -118,7 +118,7 @@ export class SerialPortDevice extends SerialPort {
   }
 
   private createRequest({ options, ...data }: SerialPortRequest) {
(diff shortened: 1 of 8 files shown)


libs/app-serialport/devices/src/lib/serial-port-device.ts
```diff
@@ -118,7 +118,7 @@ export class SerialPortDevice extends SerialPort {
   }
 
   private createRequest({ options, ...data }: SerialPortRequest) {
-    const id = parseInt(uniqueId())
+    const id = parseInt(uniqueId()) % 99999999 || 1
```
dkarski, line 121: ⚪ 
I understand this line is meant to constrain the ID range. I suggest extracting the ranged ID generation into a dedicated helper or using a single, proven library (e.g. nanoid) so the entire project relies on one consistent method.

PR #2542, opened 2025-06-26
[CP-3636] Implement AppResult for standardized response wrapping

dkarski:
JIRA Reference: [CP-3636]

### :memo: Description ️

-

[CP-3636]: https://appnroll.atlassian.net/browse/CP-3636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/app-utils/main/src/lib/app-http/app-http.service.ts
+++ b/libs/app-utils/main/src/lib/app-http/app-http.service.ts
@@ -3,30 +3,53 @@
  * For licensing, see https://github.com/mudita/mudita-center/blob/master/LICENSE.md
  */
 
-import axios from "axios"
-import { AppHttpRequestConfig, AppHttpResponse } from "app-utils/models"
+import axios, { AxiosError } from "axios"
+import {
+  AppError,
+  AppErrorType,
+  AppHttpFailedResult,
+  AppHttpRequestConfig,
+  AppHttpResult,
+} from "app-utils/models"
 
-// TODO: Implement AppError (or ObjectResult pattern)
-const isError = (error: unknown): error is Error => {
-  return error !== null && typeof error === "object" && "type" in error
-}
-
-const mapToError = (error: unknown): Error => {
-  if (isError(error)) {
-    return error
-  }
-  return new Error("Unknown error occurred")
+// TODO: Candidate for a utility function
+const getErrorMessage = (error: unknown): string => {
+  if (typeof error === "string") return error
+  if (error instanceof Error) return error.message
+  return String(error)
 }
 
 export class AppHttpService {
-  async request<T = unknown>(
+  static async request<Data = unknown>(
     config: AppHttpRequestConfig
-  ): Promise<AppHttpResponse<T> | Error> {
+  ): Promise<AppHttpResult<Data>> {
     try {
-      const { data, status } = await axios.request<T>(config)
-      return { data, status }
+      const { data, status } = await axios.request<Data>(config)
+      return { ok: true, data, status }
     } catch (error) {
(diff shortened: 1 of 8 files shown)


libs/app-utils/main/src/lib/app-http/app-http.service.ts
```diff
@@ -3,30 +3,53 @@
...
+    }
+  }
+  private static isAxiosError(error: unknown): error is AxiosError {
+    return (
+      typeof error === "object" &&
+      error !== null &&
+      "isAxiosError" in error &&
+      (error as any).isAxiosError === true
+    );
+  }
```
mkurczewski, line 40: There is already an `isAxiosError` method available in `axios` package.

libs/app-utils/main/src/lib/app-http/app-http.service.ts
```diff
@@ -3,30 +3,53 @@
...
+  ): AppHttpFailedResult<AppErrorType, ErrorData> {
+    if (this.isAxiosError(error)) {
+      return {
+        ok: false,
+        error: new AppError(error.message),
+        data: error.response?.data as ErrorData,
+        status: error.response?.status,
+      };
     }
+    return { ok: false, error: new AppError(getErrorMessage(error)) };
```
mkurczewski, line 53: `getErrorMessage` returns a message from Error, but the first argument of AppError is `type`

libs/app-utils/models/src/lib/app-error.ts
```diff
@@ -0,0 +1,26 @@
...
+export class AppError<
+  Type extends AppErrorType = AppErrorType,
+  Payload = unknown,
+> extends Error {
+  override message: string
+
+  constructor(
+    public readonly type: Type = "UnknownError" as Type,
+    message = "",
+    public readonly payload?: Payload
```
mkurczewski, line 17: Unused field

libs/app-utils/models/src/lib/app-error.ts
```diff
@@ -0,0 +1,26 @@
...
+ * For licensing, see https://github.com/mudita/mudita-center/blob/master/LICENSE.md
+ */
+
+export type AppErrorType = string
+
+export class AppError<
+  Type extends AppErrorType = AppErrorType,
+  Payload = unknown,
+> extends Error {
+  override message: string
```
mkurczewski, line 12: Error class already has the `message` field so why override it? Shouldn't this be enough?

```TSX
export class AppError<
  Type extends AppErrorType = AppErrorType,
  Payload = unknown,
> extends Error {
  constructor(
    public readonly type: Type = "UnknownError" as Type,
    message = "",
    public readonly payload?: Payload
  ) {
    super(message)

    if (Error.captureStackTrace) {
      Error.captureStackTrace(this, AppError)
    }
  }
```

libs/app-utils/models/src/lib/app-result.ts
```diff
@@ -0,0 +1,35 @@
...
+    return { ok: true, data }
+  },
+
+  failed<ErrorType extends AppErrorType = AppErrorType, ErrorData = unknown>(
+    error: AppError<ErrorType>,
+    data?: ErrorData
+  ): AppFailedResult<ErrorType, ErrorData> {
+    return { ok: false, error, data }
+  },
+}
```
mkurczewski, line 35: Both `AppResult` and `AppResultFactory` are unused

mkurczewski (approved):

